PR #6, opened 2019-07-19
add basic CI linting (no unit or acceptance tests)

jhoblitt:


Gemfile
```diff
@@ -0,0 +1,71 @@
...
+    ['>= 0', { path: File.expand_path(file_url[:path]), require: false }]
+  else
+    [place_or_version, { require: false }]
+  end
+end
+
+ruby_version_segments = Gem::Version.new(RUBY_VERSION.dup).segments
+minor_version = ruby_version_segments[0..1].join('.')
+
+group :development do
```
womullan: Why are these gems special ?

jhoblitt: This file is boilerplate emitted by the pdk (puppet development kit) for bootstrapping or converting an existing puppet mod.  Using named groups of gems is more or less a bundler best practice and its pretty common to have gems present for development that should be excluded from a production deploy, acceptance tests, etc. -- I don't think it makes any practice difference in this case as there is not a separate group for acceptance testing.

There is voodoo going on in this block to make the setup work across multiple ruby versions and on different platforms.  The goal is to pull in the correct `puppet-module-*` gem, which is essentially a meta package that pulls in a bunch of development tools.

PR #7, opened 2019-07-22
add r10k travis linting

jhoblitt:


womullan:
we need @hreinking  to look at these - the last change apparently caused a problem on the mountain.

jhoblitt:
This is a linting only config and shouldn't be able to break an env.

 @hreinking What did the pinning of puppet module versions break? I literally copied the version strings from a fresh r10k deployment. The versions should literally be the same as what you would get running r10k with the previous `Puppetfile` as run last Friday.

jhoblitt:
I'm going to go ahead and merge this but I would still be interested to hear about any breakage that was caused by the mod pinning.

PR #9, opened 2019-11-09
rm profile::it::ims (no longer used)

jhoblitt:


adrienthebo:
I can't review the functional correctness of this change at the moment but it looks reasonable. Soft :+1:, will switch to hard :+1: and merge upon verification.

PR #24, opened 2020-02-04
IT-1706: switch common layer to org layer

adrienthebo:
This commit switches the Hiera hierarchy to use organizational defaults
as the global layer, rather than a shared `common.yaml`. This change
ensures that we're deliberate about the scope of shared configuration
settings across NCSA and LSST.


andylytical:
Needs `common.yaml` added back in to the hiera hierarchy (per discussion from https://github.com/lsst-it/lsst-puppet-hiera/pull/55)

PR #25, opened 2020-02-06
IT-1686: add NFS module, nfs profile

adrienthebo:
This pull request adds an NFS module for managing nfs exports and mounts, and
adds in some shims in our profile module to define NFS exports and mounts in
hiera.

I propose that we fork the derdanne/nfs module, apply some fixups, and inject
the hiera resource creation directly into the module. We can push these changes
upstream.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -38,6 +38,7 @@ mod 'puppetlabs/mailalias_core', '1.0.5'
 mod 'puppetlabs/docker', '3.8.0'
 mod 'puppetlabs/reboot', '2.2.0'
 mod 'puppet/python', '3.0.1'
+mod 'derdanne/nfs', '2.1.2'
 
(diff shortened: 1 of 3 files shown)


Puppetfile
```diff
@@ -38,6 +38,8 @@ mod 'puppetlabs/mailalias_core', '1.0.5'
 mod 'puppetlabs/docker', '3.8.0'
 mod 'puppetlabs/reboot', '2.2.0'
 mod 'puppet/python', '3.0.1'
+mod 'derdanne/nfs', '2.1.2'
+mod 'puppet-posix_acl', '1.0.0'
```
jhoblitt: This is a nit-pick but I would like to consistently use `/` or `-` as the separator.

adrienthebo: I've reverted this commit, please take another look.

PR #31, opened 2020-02-14
add sudo rules to run commands as ccs role user

jhoblitt:


jhoblitt:
@adrienthebo I'm not following the comment?  The jira slug is part of the branch name.

PR #36, opened 2020-02-18
fix org/<org>/role/<role> paths

jhoblitt:


jhoblitt:
hiera is totally broken because of the screwd up paths.  I plan to merge this as soon as travis passes.

PR #28, opened 2020-02-11
Add NCSA support

andylytical:
Puppetfile
==========
- Use custom tuned to fix issue in custom puppet fact
- Bump version of puppet/rsyslog to 4.4.0
- Add modules needed by NCSA hosts

HIERA
=====
- Add hieradata/org/ncsa.yaml
- Add hieradata/site/npcf.yaml
- Add hieradata/site/nts.yaml

ROLES
=====
- Add "default"

PROFILES
========
- Add profile::baseline_cfg
- Add profile::system_authnz

Changes to be committed:
  modified:   Puppetfile
  new file:   hieradata/org/ncsa.yaml
  new file:   hieradata/role/default.yaml
  new file:   hieradata/site/npcf.yaml
  new file:   hieradata/site/nts.yaml
  new file:   site/profile/manifests/baseline_cfg.pp
  new file:   site/profile/manifests/system_authnz.pp

PR #42, opened 2020-02-20
IT-1778 use lsst-it tuned module

andylytical:
Move forked tuned module into lsst-it github organization.

Caveat from #28 

Fork Fixes puppet fact error message ...
```Error: Facter: error while resolving custom fact "tuned_version": undefined method `empty?' for nil:NilClass```

Also removed references to IT-1383


andylytical:
> Note that there is a mod branch change unrelated to the commit & PR title here, which would best be in its own commit.

@jhoblitt Thanks for reviewing. I don't follow you on your comment. Can you elaborate?

andylytical:
> > Note that there is a mod branch change unrelated to the commit & PR title here, which would best be in its own commit.
> 
> @jhoblitt Thanks for reviewing. I don't follow you on your comment. Can you elaborate?

Ahha, I see what you're referring to.  I'll pull that out into it's own change.

andylytical:
Pulled out the mod change for baseling_cfg into it's own branch. I'm going to merge this now.

PR #47, opened 2020-02-28
X2go testing, added to node eas-hvac01

hreinking:


Puppetfile
```diff
@@ -39,6 +39,12 @@ mod 'puppetlabs/reboot', '2.2.0'
 mod 'puppet/python', '3.0.1'
 mod 'derdanne/nfs', '2.1.2'
 
+# for mate
```
adrienthebo: Since we're not including the `mate` class, let's drop the module and the dependencies from this file.

hreinking: python module isn't being used by anything else?

PR #48, opened 2020-03-02
Add "bash completion" and "bash completion extras"

hreinking:
Add  bash completion module, and works for system native packages (such as systemctl, journalctl, ls....) but not for kubectl. In order for kubectl autocompletion to work we need to run "kubectl completion bash >/etc/bash_completion.d/kubectl". This is not complete and is going to be approach in the next PR.


jhoblitt:
This look fine to me.  All I'm really after is the kubectl tab completion for the rke user but it probably makes sense to generally enable tab completion for all users/hosts in the future.

PR #50, opened 2020-03-04
Hypervisor Role for the base

hreinking:


jhoblitt:
My preference is that tu, cp, and ls follow the same pattern. I'm fine with moving from cluster/role to site/roles for hypervisor but request that either the existing tu/cp cluster/roles be refactored at the same time or shorty there after.  I'm concerned about the cognitive load of having to sort out different sites doing essentially the same configuration but using different hiera layers.

PR #51, opened 2020-03-04
Hypervisor Role for the 3 sites

hreinking:


hieradata/site/cp/role/hypervisor.yaml
```diff
@@ -26,10 +26,10 @@ network::interfaces_hash:
     bootproto: "none"
     onboot: "yes"
     vlan: "yes"
-    bridge: "br1621"
+    bridge: "1621"
```
jhoblitt: This looks fishy...

hieradata/site/cp/role/hypervisor.yaml
```diff
@@ -26,10 +26,10 @@ network::interfaces_hash:
...
     type: "none"
     nozeroconf: "yes"
-  br1621:
+  1621:
```
jhoblitt: This also looks fishy.

PR #52, opened 2020-03-04
IT-1865 ncsa profiles for admin access

andylytical:
Summary:
- Puppetfile
  - Add supporting modules
    - inkblot/ipcalc
    - lsst/xcat
  - replace "ref" keyword with "branch"
- Move existing profiles to "ncsa" subdirectory
- Add new profiles to "profiles/ncsa"
- Add / update hiera data for new profiles


site/profile/hiera.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
adrienthebo: What part of the hiera hierarchy is touching `site/profile/hiera.yaml`?

site/profile/manifests/ncsa/allow_ssh_from_login.pp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    #add in allow_groups if non-empty
+    if size( $allow_groups ) > 0 {
+        $groups = { 'AllowGroups' => $allow_groups }
+    }
+    else {
+        $groups = {}
+    }
+    #join the param hashs
```
adrienthebo: General code review: when commenting code like this, comments should indicate why things are done, rather than what's being done. We can generally infer what the code is doing, so why are we doing the joining? Why is this behavior interesting and what should code readers know about what this means?

site/profile/manifests/ncsa/allow_ssh_from_bastion.pp
```diff
@@ -0,0 +1,45 @@
...
+#   Type: Array
+#   Desc: one or more LDAP / UNIX groups that are allowed to login from
+#         any of the nodes in bastion_nodelist
+#         Note: can be empty
+# @example
+#   include profile::allow_ssh_from_bastion
+#
+class profile::ncsa::allow_ssh_from_bastion (
+    Array[ String ]    $allow_groups,
+    Array[ String, 1 ] $bastion_nodelist,
```
jhoblitt: Is `1` allowed here?

andylytical: 1 indicates minimum size ... ie: array cannot be empty

site/profile/hiera.yaml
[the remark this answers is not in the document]
andylytical: This is module level hiera.

andylytical: This is module level hiera

site/profile/manifests/ncsa/allow_ssh_from_login.pp
andylytical: Point taken. Removed redundant comment.

site/profile/hiera.yaml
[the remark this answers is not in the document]
andylytical: This is module level hiera data.

PR #55, opened 2020-03-05
IT-1860 general docker profile

andylytical:
Create a generic docker profile
```
profile::docker
```
that can be included in any role. This is essentially a copy of:
```
hieradata/org/lsst/role/docker-compose.yaml
```
that splits the `include`s into the profile and the data into hiera.

### DIFF
--- a/hieradata/org/ncsa.yaml
+++ b/hieradata/org/ncsa.yaml
@@ -13,6 +13,13 @@ baseline_cfg::networkmanager::ensure: "running"
 
 chrony::leapsectz: "right/UTC"
 
+docker::overlay2_override_kernel_check: true
+docker::storage_driver: "overlay2"
+docker::version: "19.03.4"
+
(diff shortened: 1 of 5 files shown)


hieradata/org/ncsa.yaml
```diff
@@ -13,6 +13,13 @@ baseline_cfg::networkmanager::ensure: "running"
 
 chrony::leapsectz: "right/UTC"
 
+docker::overlay2_override_kernel_check: true
```
adrienthebo: Non-blocker: we're copy/pasting these values in a lot of place, I'd like to make them defaults. Thoughts?

andylytical: The module level hiera data for the profile would be a good place to make them defaults. I'm okay with that.
I'm happy to put those in if there is agreement.

andylytical: nevermind, it can't go in module level hiera data (as is). It would require changing the profile to take parameters and then set defaults for those parameters in module level hiera. I'll proceed with the merge as is and leave this topic for a future change.

PR #57, opened 2020-03-06
X2go module for was-hvac01

hreinking:


adrienthebo:
I have existing comments on this branch in PR #47 - where possible we should use the same pull request for work so that we can keep the context of comments.

adrienthebo:
Closed #57 in favor of #47.

PR #59, opened 2020-03-09
maint: bump dhcp1 lease duration to 86400

adrienthebo:


hieradata/cluster/corecp/role/dhcp.yaml
```diff
@@ -1,8 +1,8 @@
 ---
 dhcp::interfaces:
   - "ens192"
-dhcp::default_lease_time: 3600
-dhcp::max_lease_time: 3600
+dhcp::default_lease_time: 43200
```
jhoblitt: I think 12 hour hours is much to long.  This would mean waiting up to 12 hours for any dhcp configuration changes to to propagate.

adrienthebo: We're using 43200/86400 at Tucson; I originally bumped this when we were having issues with DHCP at the summit and intended to reconcile this value with Tucson now that DHCP has settled down in the summit. I don't feel strongly about this value, if you think there's value in keeping this lease time short then let's close out this PR.

PR #60, opened 2020-03-09
IT-1454: regenerate types when r10k deploys modules

adrienthebo:
Puppet uses abstract type definitions to prevent types from leaking between
different Puppet environments, but require an additional generation step for
them to be loadable by Puppet. This modifies r10k to perform that regeneration
when deploying an environment.

### DIFF
--- a/hieradata/org/lsst/role/foreman.yaml
+++ b/hieradata/org/lsst/role/foreman.yaml
@@ -34,3 +34,5 @@ r10k::webhook::config::protected: false
 r10k::webhook::use_mcollective: false
 r10k::webhook::user: "root"
 r10k::webhook::group: "root"
+r10k::deploy_settings:
+  generate_types: true


adrienthebo:
Sure - so the root issue is that Puppet type definitions used to leak across environments because Puppet would synthesize Ruby constants, and those wouldn't be scoped to specific environments. The new approach is that Puppet takes the Ruby type definitions and converts them into Puppet code which is properly scoped to environments. This generates the following, using elasticsearch as an example:

**`$environment/.resource_types/elasticsearch_user.pp`**:
```
# This file was automatically generated on 2020-03-09 15:30:20 +0000.
# Use the 'puppet generate types' command to regenerate this file.

# Type to model Elasticsearch users.
Puppet::Resource::ResourceType3.new(
  'elasticsearch_user',
  [
    # The basic property that the resource should be in.
    #
    # Valid values are `present`, `absent`.
    Puppet::Resource::Param(Enum['present', 'absent'], 'ensure')
  ],
  [
    # User name.
    Puppet::Resource::Param(Any, 'name', true),

    # Path to the elasticsearch configuration directory (ES_PATH_CONF).
    Puppet::Resource::Param(Any, 'configdir'),

    # Plaintext password for user.
    #
    #
    #
    # Requires features manages_plaintext_passwords.
    Puppet::Resource::Param(Any, 'password'),
```

So the Ruby type definition is no longer used. https://puppet.com/docs/puppet/latest/environment_isolation.html has the full story for what's going on here.

PR #61, opened 2020-03-09
add antu cluster

jhoblitt:


hieradata/cluster/antu.yaml
```diff
@@ -0,0 +1,35 @@
...
+    defroute: "yes"
+    onboot: "yes"
+    type: "Ethernet"
+  em2:
+    bootproto: "none"
+    onboot: "no"
+    type: "Ethernet"
+yum::managed_repos:
+  - "dell"
+yum::repos:
```
adrienthebo: We have a number of settings that we apply for Dell hardware, and this configuration isn't specific to/particularly fundamental for the antu cluster. I'd like to extract a `dell_server` class for details like this - what's your thinking on that approach?

hieradata/cluster/antu.yaml
```diff
@@ -0,0 +1,35 @@
+---
+lookup_options:
+  network::interfaces_hash:
+    merge:
+      strategy: "deep"
+  yum::managed_repos:
```
adrienthebo: Why is this unique?

hieradata/cluster/antu.yaml
jhoblitt: The only dell specific resource is the installation of `perccli` and the associated yum repo.  I do not object to a perccli profile being created so long as the **data** stays in yaml.

hieradata/cluster/antu.yaml
jhoblitt: As otherwise  conflicting yum repo names would attempt to be merged.  I'd rather get one out of 2 repos rather than a total trainwreck.

jhoblitt: (although I really wish a collision would be fatal...)

PR #65, opened 2020-03-13
IT-1480: reimplement telegraf

adrienthebo:
This pull request reimplements telegraf as a monitoring agent on Puppet hosts.

TODO:

* [x] Ensure that this change won't impact the summit monitoring servers
* [x] Figure out how we'll selectively enable this on sites that have influx servers

### DIFF
--- a/site/profile/manifests/core/common.pp
+++ b/site/profile/manifests/core/common.pp
@@ -1,3 +1,5 @@
+# @summary
+#   Common functionality needed by standard nodes.
 class profile::core::common {
   include timezone
   include tuned
--- a/site/profile/manifests/core/common.pp
+++ b/site/profile/manifests/core/common.pp
@@ -1,6 +1,13 @@
 # @summary
 #   Common functionality needed by standard nodes.
-class profile::core::common {
+#
+# @param collect_metrics
+#   Enable or disable metrics collection. Metrics collection may be disabled on development
+#   nodes, nodes that don't have uptime requirements, or nodes that should only have minimal
+#   software load.
+class profile::core::common(
+  Boolean $collect_metrics = true,
+) {
   include timezone
   include tuned
   include chrony
@@ -18,5 +25,8 @@ class profile::core::common {
   include augeas
   include rsyslog
   include rsyslog::config
-  include profile::core::telegraf
+
+  if $collect_metrics {
+    include profile::core::telegraf
+  }
 }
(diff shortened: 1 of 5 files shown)


adrienthebo:
Testing comment notifications for pull requests.

site/profile/manifests/core/telegraf.pp
```diff
@@ -0,0 +1,60 @@
...
+  String $host     = 'metrics-ingress.ls.lsst.org',
+) {
+  class { '::telegraf':
+    hostname    => $::facts['fqdn'],
+    global_tags => {'site' => $::site},
+  }
+
+  $default_inputs = {
+    'cpu'        => [{}],
+    'chrony'     => [{}],
```
adrienthebo: It might be reasonable to slice this out to a `profile::core::chrony` module, but it's harmless as-is.

site/profile/manifests/core/common.pp
```diff
@@ -8,12 +17,15 @@
   include sysstat
   include epel
   include sudo
-  include accounts
```
jhoblitt: removing the `accounts` mod here without re-adding it elsewhere will cause a massive amount of breakage.

site/profile/manifests/core/common.pp
```diff
@@ -1,4 +1,13 @@
-class profile::core::common {
+# @summary
+#   Common functionality needed by standard nodes.
+#
+# @param collect_metrics
+#   Enable or disable metrics collection. Metrics collection may be disabled on development
+#   nodes, nodes that don't have uptime requirements, or nodes that should only have minimal
+#   software load.
+class profile::core::common(
+  Boolean $collect_metrics = true,
```
jhoblitt: I would prefer this to be a direct include of the `profile::core::telegraf` class in hiera as it's one line in hiera either way and direct inclusion involves less indirection. However, I can live with it.

site/profile/manifests/core/common.pp
adrienthebo: Good catch, this came from an aborted refactor.

site/profile/manifests/core/common.pp
adrienthebo: Let's break this out into a conversation on classification and roles and profiles, are you available to discuss this at tomorrow's Configuration Management meeting?

PR #68, opened 2020-03-16
Missing class in previuos PR

hreinking:


hreinking:
> I'm blocking this pull request because our copy/pasting of module code is out of control. @hreinking stick with local development on a topic branch for now, but let's not merge this.

This is a first commit; I'm planning to add more packages in the future and test the performance, without affecting the hypervisor role.

hieradata/org/lsst/role/dev_hyp.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
adrienthebo: If this is extending the hypervisor class, then add this as an override, either at the node level (eg `hieradata/nodes/core-dev.ls.lsst.org`), or add this node to a cluster and configure the cluster there (eg `hieradata/clusters/coredev/role/hypervisor.yaml`. Copy/pasting configuration like this is harmful because it allows for slight but dangerous inconsistencies in our Puppet code.

If you're doing experimental work, you can do so in a branch - only code that's relatively ready for production needs to be merged.

PR #79, opened 2020-03-30
add antu em2.2400 interface

jhoblitt:


hieradata/cluster/antu.yaml
```diff
@@ -23,6 +23,13 @@ network::interfaces_hash:
     bootproto: "none"
     onboot: "no"
     type: "Ethernet"
+  em2.2400:  # 139.229.147.0/24
```
adrienthebo: Are we going to run into routing issues like we saw with the dual-homed summit hosts?

jhoblitt: Yes but as the summit is currently down for months, there's no way of testing the routing so I intentionally decided to omit the routes.

PR #80, opened 2020-03-30
maint: fixup telegraf configuration

adrienthebo:
This commit renames the telegraf user and database to match the defaults, and fixes up some bad outputs.

Note that there's a matching private Hiera PR.

### DIFF
--- a/site/profile/manifests/core/telegraf.pp
+++ b/site/profile/manifests/core/telegraf.pp
@@ -14,8 +14,8 @@
 #   The backing InfluxDB instance. At present the default DNS record is a CNAME.
 class profile::core::telegraf(
   String $password,
-  String $username = 'telegraf-nodes',
-  String $database = 'nodes',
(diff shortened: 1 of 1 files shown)


site/profile/manifests/core/telegraf.pp
```diff
@@ -14,13 +14,14 @@
...
-  String $username = 'telegraf-nodes',
-  String $database = 'nodes',
+  String $username = 'telegraf',
+  String $database = 'telegraf',
   String $host     = 'metrics-ingress.ls.lsst.org',
 ) {
   class { '::telegraf':
     hostname    => $::facts['fqdn'],
     global_tags => {'site' => $::site},
+    outputs     => {},  # suppress class defaults, it populates an empty influx instance
```
adrienthebo, line 24: It might be more reasonable to use this output rather than the separate `telegraf::output` defined type.

PR #84, opened 2020-04-02
add comcam network

hreinking:


hieradata/site/ls/role/foreman.yaml
```diff
@@ -40,6 +40,12 @@ dhcp::pools:
     gateway: "139.229.147.254"
     range: "139.229.147.225 139.229.147.253"  # ~ /27
     search_domains: "%{alias('dhcp::dnsdomain')}"
+  ComCam:
+    network: "139.229.150.0"
+    mask: "255.255.255.128"
+    gateway: "139.229.150.127"
+    range: "139.229.150.1 139.229.150.100"
```
jhoblitt, line 47: This range is probably much to large.  I would suggest cutting it down to the upper /27 or /28.

jhoblitt:
I've made a couple of changes:

* I've renamed the subnet from comcam -> comcam-ccs
* changed the gw from .127 (broadcast) to .126 (I'm assuming this is the gateway)
* I've changed the DHCP pool range to be .112 - .125 which is the upper ~ /28 of the subnet

PR #85, opened 2020-04-02
add base archive network dhcp

jhoblitt:

PR #86, opened 2020-04-03
add base archive network dhcp

jhoblitt:


jhoblitt:
This is a reopening of #85 which was very oddly not updating with changes to its source branch.

PR #93, opened 2020-04-07
IT-2115: implement bastion role and remote_desktop profile

adrienthebo:


adrienthebo:
Renamed from `shell_host` to `bastion`.

PR #106, opened 2020-04-22
Update ncsa ldap group names

andylytical:
Admin group was renamed.
Remove duplicate group from sssd (since it's added by sshd)


andylytical:
FYI @adrienthebo @jhoblitt 
I don't think any review is necessary, changed LDAP group names in hieradata/org/ncsa.yaml
I plan to merge once Travis checks are complete.

PR #110, opened 2020-04-24
ifdown class

hreinking:
I couldn't find any class that completely pulls down the link, so I wrote the class ifdown. Please let me know your thoughts.


site/profile/manifests/core/ifdown.pp
```diff
@@ -0,0 +1,17 @@
+# @summary
+#   Special case for servers that do PXE from one interface and change
+#   to another one after the first puppet run  IT-2199
+
+class profile::core::ifdown (
+  $interface,
+)
+{
+  $command = "ifconfig ${interface} down"
```
jhoblitt, line 9: It would be safer to use the `ip` command as `ifconfig` isn't part of the el7 @core.  I just happen to have added it to the rke role.

hreinking: changed to ip instead of ifconfig

PR #117, opened 2020-04-29
manage lsst-daq interface name on dhcp client nodes

jhoblitt:


jhoblitt:
Repushed with a whitespace fix (I'm surprised the linter didn't catch it).

PR #118, opened 2020-04-29
rm lsst/ccs & lsst/headerservice mods (unused)

jhoblitt:


jhoblitt:
neither of these mods is used in the current code tree and both of the repos have been marked as archived.

PR #123, opened 2020-05-01
add profile::ccs::sysctl

jhoblitt:


jhoblitt:
@glennmorris I think this would probably be better as part of `ccs_profile` or other external module but I didn't want to pin down the refactoring you've been doing while learning.  I think we should probably start a new module next week.

PR #124, opened 2020-05-05
remove hardware specific network config from top level roles

jhoblitt:


hieradata/site/cp/role/hypervisor.yaml
```diff
@@ -3,6 +3,8 @@ lookup_options:
   accounts::user_list:
     merge:
       strategy: "deep"
+classes:
```
adrienthebo, line 6: Is this reasonable to push to the top level hypervisor role? Is there a scenario where we do want a hypervisor but we don't want to manage the network?

jhoblitt: This PR is doing the opposite and pushing it from the top level role down to the site because this is where the `network::interfaces_hash` key first appears.  I don't have strong feelings about it either way.

PR #127, opened 2020-05-11
Point telegraf at the new influxdb CNAME

adrienthebo:


site/profile/manifests/core/telegraf.pp
```diff
@@ -10,15 +10,14 @@
...
-# @param host
+# @param influxdb_url
 #   The backing InfluxDB instance. At present the default DNS record is a CNAME.
 class profile::core::telegraf(
   String $password,
-  String $username = 'telegraf',
-  String $database = 'telegraf',
-  String $host     = 'metrics-ingress.ls.lsst.org',
+  String $username     = 'telegraf',
+  String $database     = 'telegraf',
```
hreinking, line 18: there's no concern on leaving the name of the db and the user in plain text?

adrienthebo: These are default values and don't really need to be secured; the password is safely stored and should provide sufficient security for our needs.

PR #134, opened 2020-05-18
dtn proposal role

hreinking:
ls site and lsst org will be to invasive for Ampath's dtn, so I created this node role. Looking forward for comments (if I'm missing something or the role has to much on it). I'm also looking a way to disable foreman's org and site.


csilva-cl:
role seems good to me.

PR #139, opened 2020-05-20
It 2277 ruka04

hreinking:
Successfully tested fixing the kernel version for the new node, and also added the long ago requested kubectl completion feature.


hieradata/node/ruka04.ls.lsst.org.yaml
```diff
@@ -0,0 +1,2 @@
+---
+profile::core::kernel::version: "3.10.0-1127.8.2.el7.x86_64"
```
adrienthebo, line 2: In `hieradata/cluster/ruka.yaml` There's an explicit kernel version applied to the cluster. Since ruka04 is part of the cluster is this setting needed?

hreinking: not going to use that cluster definition, cause I'm going to use an Antu node.

hreinking: I'm going to use the antu cluster definition, and that one doesn't have the kernel patch you applied

hreinking:
trying to unhang the ci

PR #144, opened 2020-05-27
cleanup hypervisor role + setup ls-fwd01 as comcam ccs hypervisor

jhoblitt:


hieradata/node/core1.dev.lsst.org.yaml
```diff
@@ -0,0 +1,24 @@
+---
```
hreinking, line 1: This node definition isn't needed, cause it uses the same as the core nodes

hreinking: Not needed and it will crush core1.dev, the interface should be p2p1.2101, and managing p2p1 instead of p2p2

hieradata/node/core2.dev.lsst.org.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
hreinking, line 1: lgtm

hieradata/node/core1.dev.lsst.org.yaml
jhoblitt: The vlan id isn't the same as the core.ls nodes (2101 vs 2102).  I will check/fix the interface name.

jhoblitt: I've fixed the trunk interface name on core1.dev and also removed manually removed the non-working br1800, p2p1.1800 and p2p1.2101 interfaces.

jhoblitt:
The hiera cleanups have been split out as a separate PR: #150

PR #145, opened 2020-05-28
telegraf: reduce load on influxdb

adrienthebo:


adrienthebo:
> I'm guessing this means up to 20s of data could be lost in a crash? I'm certainly OK with that.

Correct, but we have no uptime requirements for telegraf yet so this is acceptable at present; if we receive a new requirement for higher durability then we can revise this downward again.

PR #169, opened 2020-07-01
 add lhn static routes to comcam-frwd02 

jhoblitt:


jhoblitt:
This PR is only duplicating config comcam-arctl01 to associated nodes so I'm going to skip the review process in the interest of time.

PR #177, opened 2020-07-07
 rm dead roles (graylog, influxdb, grafana)

jhoblitt:


jhoblitt:
rebased on current master

PR #199, opened 2020-08-07
puppetdb + basic puppetserver/foreman management

jhoblitt:
Note that foreman-proxy management is not included.


jhoblitt:
Related secrets: https://github.com/lsst-it/lsst-puppet-hiera-private/pull/29

jhoblitt:
I've revised this PR so that puppetdb support will not be enabled for CP by default.

PR #200, opened 2020-08-10
adding boolean for install, not installing in tu.

hreinking:
quick fixup to solve adding dev and setting tuc to false

### DIFF
--- a/hieradata/site/dev.yaml
+++ b/hieradata/site/dev.yaml
@@ -105,3 +105,7 @@ accounts::user_list:
 firewall::ensure: "stopped"
 
 letsencrypt::server: "https://acme-staging.api.letsencrypt.org/directory"  # testing url
+profile::core::common::deploy_icinga_agent: false
+profile::core::icinga_agent::icinga_master_fqdn: "icinga-master.ls.lsst.org"
+profile::core::icinga_agent::icinga_master_ip: "139.229.135.31"
+profile::core::icinga_resources::dhcp_server: "139.229.134.5"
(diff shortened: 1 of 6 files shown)


hieradata/site/dev.yaml
```diff
@@ -105,3 +105,7 @@ accounts::user_list:
 firewall::ensure: "stopped"
 
 letsencrypt::server: "https://acme-staging.api.letsencrypt.org/directory"  # testing url
+profile::core::icinga_agent::install: true
+profile::core::icinga_agent::icinga_master_fqdn: "icinga-master.ls.lsst.org"
+profile::core::icinga_agent::icinga_master_ip: "139.229.135.31"
+profile::core::icinga_resources::dhcp_server: "139.229.135.5"
```
jhoblitt, line 111: Isn't this the wrong dhcp server for dev?

site/profile/manifests/core/icinga_agent.pp
```diff
@@ -2,95 +2,99 @@
 #   Icinga agent creation for metric collections
 
 class profile::core::icinga_agent(
+  Boolean $install,
```
jhoblitt, line 5: I don't like this approach. It would be better to add a boolean param to `profile::core:common` as I suggested in review to turn inclusion of this class on and off.  The advantage of that is only needing one key to turn the agent behavior on and off instead of having to define a bunch of unused keys per site.

hieradata/site/dev.yaml
hreinking: fixed

site/profile/manifests/core/icinga_agent.pp
hreinking: done!

PR #201, opened 2020-08-10
Setting icinga-agent flag to enable in dev site

hreinking:


jhoblitt:
What is being enabled?

hreinking:
@jhoblitt I accidentally left dev in disable

PR #202, opened 2020-08-11
bypass pihole and resolve directly to AD

jhoblitt:


jhoblitt:
Could someone confirm that this is the correct IP for AD resolution?  There is a resolver at that IP.

PR #203, opened 2020-08-11
rm pihole role

jhoblitt:


jhoblitt:
@andylytical I'm guessing you won't object to removing the `puppet/firewalld` mod?

andylytical:
@jhoblitt No objections to removing `puppet/firewalld`. Thanks for asking!

PR #204, opened 2020-08-11
dedup inclusion of sysctl::values

jhoblitt:


jhoblitt:
Sysctls are already broadly used.  Once we globally disable ipv6, this mod will be in use on all of our nodes.

hreinking:
Then it sounds goo to me

PR #205, opened 2020-08-11
dedup inclusion of network

jhoblitt:


jhoblitt:
The vast majority of nodes have this class included either through a cluster or node layer.  It is so widely used that we should just assume that its always present.

hreinking:
Ok, your call.

jhoblitt:
You are correct that this may result in a some nodes have a module in their manifest which is unused and that is a situation that general should be avoided.  However, we are so widely using this module we should be able to just assume its presence rather than having dozens of places in which the classes is being included.

PR #214, opened 2020-08-13
IT-2427: icinga master as CA for icinga agents

hreinking:
Now we can add an agent that doesn't belong to puppet (like switches or vmware clusters)

### DIFF
--- a/site/profile/manifests/core/icinga_agent.pp
+++ b/site/profile/manifests/core/icinga_agent.pp
@@ -6,6 +6,7 @@ class profile::core::icinga_agent(
   String $icinga_master_ip,
   String $credentials_hash,
   String $host_template,
+  String $ca_salt,
 ){
   $packages = [
     'nagios-plugins-all',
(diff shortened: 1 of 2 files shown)


csilva-cl:
what happens with the agents? do they have to load the new cert ?

hreinking:
yes, they renegociate

hreinking:
tested with icinga-agent.ls.lsst.org

site/profile/manifests/core/icinga_master.pp
```diff
@@ -180,6 +183,8 @@
       },
     },
   }
+  include ::icinga2::pki::ca
```
jhoblitt, line 186: I presume this results in a new CA being dynamically generated?  What happens if the master is reinstalled?  Will all the agents break?

site/profile/manifests/core/icinga_agent.pp
```diff
@@ -63,10 +64,11 @@
   }
   ##Icinga2 feature API config
   class { '::icinga2::feature::api':
-    pki             => 'puppet',
```
jhoblitt, line 66: The default value for pki is "icinga2" but your setting it to `none` in the icinga_master profile, are they intentionally different values?

 https://github.com/Icinga/puppet-icinga2/blob/master/manifests/feature/api.pp#L149

site/profile/manifests/core/icinga_master.pp
csilva-cl: I think you are right, they should break, but as Heinrich said they should renogotiate

hreinking: if the master is reinstall, all agents renegociate

site/profile/manifests/core/icinga_agent.pp
hreinking: this was the original recommendation from the repo, but couldn't make it work earlier. The master identify itself as CA.

PR #223, opened 2020-08-20
IT-2373: velero client and autocompletion

hreinking:
The exec is due to no existence of a module that incorporates velero autocompletion (as well as happened with kubectl)

### DIFF
--- a/hieradata/cluster/ruka.yaml
+++ b/hieradata/cluster/ruka.yaml
@@ -1,6 +1,7 @@
 ---
 classes:
   - "profile::core::ifdown"
+  - "profile::core::velero"
 packages:
   - "perccli"  # from dell repo
 profile::core::ifdown::interface: "em1"
--- a/site/profile/manifests/core/velero.pp
+++ b/site/profile/manifests/core/velero.pp
@@ -1,32 +1,20 @@
 # @summary
 #   Velero cli instalation and autocompletion
 
-class profile::core::velero
-{
-  $runif1  = 'test ! -f /usr/bin/velero'
-  $runif2  = 'test ! -f /etc/bash_completion.d/velero'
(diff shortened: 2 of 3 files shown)


site/profile/manifests/core/velero.pp
```diff
@@ -0,0 +1,20 @@
+# @summary
+#   Velero cli instalation and autocompletion
+
+class profile::core::velero {
+  $runif  = 'test ! -f /etc/bash_completion.d/velero'
+  $command = 'velero completion bash > /etc/bash_completion.d/velero'
+
+  vcsrepo { '/usr/share/velero':
```
jhoblitt, line 8: Why is installing from git preferred over the official tarball releases?

site/profile/manifests/core/velero.pp
```diff
@@ -0,0 +1,20 @@
...
+    ensure   => present,
+    provider => git,
+    source   => 'git://github.com/vmware-tanzu/velero.git',
+    revision => 'v1.4.2',
+  }
+  exec { $command:
```
jhoblitt, line 14: The problem with looking for the existence of a file means that the content will never be updated, even if the version of the application changes.  Putting `echo 'source <(velero completion bash)' ` into a profile.d file eliminates the `exec` resource and should has better forward compatibility.

hieradata/cluster/ruka.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 classes:
   - "profile::core::ifdown"
+  - "profile::core::velero"
```
jhoblitt, line 4: Is the intent to only install velero on ruka and vk8s?  If not, this should be part of the rke role.

hreinking: Is what this branch contemplates. The integration of all others is separated into several branches. I recalling you saying "small changes at the time"

PR #229, opened 2020-09-14
IT-1739: additional features monitoring

hreinking:
Adding Network devices to monitoring and dedicated services, general code cleanup, comcam-ccs cluster and hostgroup, and features such as tls cert expiration, ntp clock skew, implicit ping (ping from another device), among others


csilva-cl:
A few comments
- the service template creation (and many others) should be possible to create arrays/functions to create them, it's basically the same command repeated.
- I think it would be good to keep pagerduty (and opsgenie) on different branches and do not merge them to the master, they are just tests. 

For future PR, please split them into smaller/modular chunks, so it's easy to review and follow them.

hreinking:
@csilva-cl Thanks for the quick review and my apologies for the length of it; I add to many tickets into the same branch and the result was this. Next time, shorter PRs.

- I tried to array the services, but each of them have different variables in the vars section, and they vary in length, so I couldn't determine a pattern that replaces the content.
- Just removed PD from the current branch and add it into IT-2452_pagerduty

csilva-cl:
There are still traces of pagerduty in the PR.

The curl command is always the same, a solution could be adding the command to a function and just passing the variables that could be stored in a 2d array would be much more effective.

PR #230, opened 2020-09-16
IT-1739: additional features monitoring

hreinking:
Adding Network devices to monitoring and dedicated services, general code cleanup, comcam-ccs cluster and hostgroup, and features such as tls cert expiration, ntp clock skew, implicit ping (ping from another device), among others


csilva-cl:
Thanks this is much easier to follow, however there's still some repetitive code that could be synthetized.

PR #232, opened 2020-10-07
IT-2472: mrtg to icinga

hreinking:
Add all checks that currently mrtg is doing to comcam hosts to icinga instance.


csilva-cl:
everything looks fine, but please, do not make another PR with so many commits if they are not part of the same component, it is quite difficult to follow.

PR #233, opened 2020-10-13
IT-2484: Summit Network Normalization

hreinking:
Site modifications, vlans relocation, enable monitoring and syslog forwarding


csilva-cl:
missing ls.lsst.org domain in the search

hreinking:
> missing ls.lsst.org domain in the search

As you may see in other site configurations, we are not including them in the search scope; this is only the dns domain, it's still going to resolve {cp,tu,ls}.lsst.org.

csilva-cl:
ok, we can work on normalization later but we should have full search to avoid fqdn

PR #255, opened 2020-11-17
comcam ls -> cp migration

jhoblitt:


jhoblitt:
https://www.traviscistatus.com/ doesn't list any problems but there is definitely some sort of travis<->github breakdown... I'm going to merge before CI has run.

PR #263, opened 2020-11-20
IT-2581: m2-control node definition for VIPM

hreinking:


csilva-cl:
you are missing ensure_packages to check that they are not present already.

hreinking:
In the current puppet release, all packages are managed through the package resource:
https://puppet.com/docs/puppet/7.0/types/package.html#package-attribute-ensure

PR #265, opened 2020-11-24
Install some more basic OS rpms on CCS nodes

glennmorris:


glennmorris:
This was premature; sorry for the noise.

PR #267, opened 2020-11-30
IT-2649: move role dtn to ls, use common profile and add flag to firewall

hreinking:


csilva-cl:
what's the objective of removing the top level dtn role ?

hreinking:
That is completely open for debate, don't have strong feelings regarding that topic. Just think that that role is particulary design for that dtn; if in the future we have another one in another site (like cp), it won't be usable.

PR #272, opened 2020-12-08
audit summit host network configuration

jhoblitt:
* The physical interfaces of all hosts (and some VMs) are now managed by puppet
* All dual-homed hosts now have their FQDN interface in a non-dds subnet
* All dual-homed hosts with a dds interface now have the appropriate static routes to keep dds traffic on the dds dedicated interfaces
* NFS exports (and mounts) have been revised
* Most hosts are now using DHCP at least for the FQDN interface
* A few critical hosts and VMs have been switch to static IPs in order to allow recovery from DNS/DHCP breakage
* ipv6 was disabled on all hosts in #271 
* Misc. cleanups


srp3rd (changes requested):
I looked at the NFS exports.  The machine where the atarchiver and ccarchiver services are running currently (at NTS and base) use the /data/lsstdata/{base|NTS}/{comcam|auxTel} directories.   I believe based on the discussion in Slack, we can make these /data/lsstdata/summit/{comCam|auxTel}.

The "staging" directories for the data buffer manager and OODS are located in /data/staging, and don't need to be exported elsewhere.  (Although, if /data/lsstdata were exported to the Forwarder machines, we could do a "cp" or "mv" instead of a "scp").  

The butler repos (both gen2 and gen3) are below the /data/lsstdata/summit/{comCam|auxTel}.   Currently on comcam-arctl01, the gen2 directory is: /data/lsstdata/base/comcam/oods/butler and the gen3 directory is /data/lsstdata/base/comcam/oods/gen3butler, but we can name those whatever we want.

srp3rd (approved):

PR #296, opened 2021-02-03
It 2751/hvac01 node

csilva-cl:
Configuration of hvac01 node + x2go agent profile by Heinrich

### DIFF
--- a/hieradata/node/bastion1.cp.lsst.org.yaml
+++ b/hieradata/node/bastion1.cp.lsst.org.yaml
@@ -1,3 +1,0 @@
----
-classes:
-  - "profile::core::x2go_agent"
--- a/hieradata/node/bastion1.cp.lsst.org.yaml
+++ b/hieradata/node/bastion1.cp.lsst.org.yaml
@@ -1,3 +1,3 @@
 ---
 classes:
-  - "x2go"
+  - "profile::core::x2go_agent"
--- a/hieradata/node/net-dx.cp.lsst.org.yaml
+++ b/hieradata/node/net-dx.cp.lsst.org.yaml
@@ -2,6 +2,7 @@
 classes:
   - "profile::core::common"
   - "profile::core::net_dx"
+  - "profile::core::x2go_agent"
 network::interfaces_hash:
   ens192:  # fqdn
     bootproto: "dhcp"
--- a/hieradata/org/lsst/role/hexrot.yaml
+++ b/hieradata/org/lsst/role/hexrot.yaml
@@ -5,7 +5,7 @@ classes:
   - "profile::core::ni_packages"
   - "docker"
   - "python"
-  - "x2go"
+  - "profile::core::x2go_agent"
 # disabling the kernel version check is needed on el7
 docker::overlay2_override_kernel_check: true
 # / on hexrot.cp was formated with xfs fstype=0 (long, long, long ago) and is
--- a/site/profile/manifests/core/common.pp
+++ b/site/profile/manifests/core/common.pp
@@ -41,7 +41,6 @@ class profile::core::common(
   Boolean $manage_ipa = true,
   Boolean $disable_ipv6 = false,
   Boolean $manage_firewall = true,
-  Boolean $install_x2go = false,
 ) {
   include accounts
   include augeas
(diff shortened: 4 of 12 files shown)


site/profile/manifests/core/common.pp
```diff
@@ -41,6 +41,7 @@
   Boolean $manage_ipa = true,
   Boolean $disable_ipv6 = false,
   Boolean $manage_firewall = true,
+  Boolean $install_x2go = true,
```
jhoblitt, line 44: `profile::core::common` is intended to be included by all roles so by setting `true` here, x2go will be installed on all hosts, is that the intent?

hieradata/node/bastion1.cp.lsst.org.yaml
```diff
@@ -1,3 +1,3 @@
 ---
 classes:
-  - "x2go"
+  - "profile::core::x2go_agent"
```
jhoblitt, line 3: This probably isn't needed assuming a node has a proper role assigned to it. This is a case where what's in the tree is an anti-example.  The `x2go` should never have been included directly at the node level and instead brought in via a role.

hieradata/node/net-dx.cp.lsst.org.yaml
```diff
@@ -2,6 +2,7 @@
 classes:
   - "profile::core::common"
   - "profile::core::net_dx"
+  - "profile::core::x2go_agent"
```
jhoblitt, line 5: There is something strange going on with this host in that `profile::core::common` should be part of the host's role and not directly included at the node level.

site/profile/manifests/core/x2go_agent.pp
```diff
@@ -0,0 +1,20 @@
+# @summary
+#   Adds additional packages for x2go agent
```
jhoblitt, line 2: This is a trivial class, and we other examples like this in the tree, but FYI, I have been trying to slowly remove them.  In general, application specific class that aren't "composing" should be an external module in the `Puppetfile`.

site/profile/manifests/core/x2go_agent.pp
```diff
@@ -0,0 +1,20 @@
...
+
+  $packages = [
+    'x2goserver',
+    'x2goclient',
+    'x2goserver-common',
+    'x2goserver-xsession',
+    'x2goagent',
+  ]
+
+  package { $packages:
```
jhoblitt, line 14: It is good practice to use the `ensure_packages()` function to prevent resource conflicts with other classes.

site/profile/manifests/core/x2go_agent.pp
```diff
@@ -0,0 +1,20 @@
...
+    ensure => 'present',
+  }
+  yum::group { 'Mate':
```
jhoblitt, line 17: I see that there is a `mate` mod in the `Puppetfile`. Can we use that to install the mate desktop?

jhoblitt (changes requested):

site/profile/manifests/core/common.pp
csilva-cl: No, that's not. I will turn it off. Thanks.

site/profile/manifests/core/x2go_agent.pp
csilva-cl: ok

site/profile/manifests/core/x2go_agent.pp
csilva-cl: I'll give it a go.

site/profile/manifests/core/x2go_agent.pp
csilva-cl: Noted.

hieradata/node/net-dx.cp.lsst.org.yaml
csilva-cl: yes I agree

hreinking (approved):
LGTM

hieradata/org/lsst/role/hexrot.yaml
```diff
@@ -5,7 +5,7 @@ classes:
   - "profile::core::ni_packages"
   - "docker"
   - "python"
-  - "x2go"
+  - "profile::core::x2go_agent"
```
jhoblitt, line 8: I'd prefer that we use `profile::core::common::install_x2go: true` or directly include the `profile::core::x2go_agent` class but try to avoid using two idioms that accomplish the same result.

jhoblitt (approved):
This looks ready to go.

PR #310, opened 2021-04-08
Comcam mcm sal

glennmorris:
This will enable management of CCS SAL on comcam-mcm, as per https://jira.lsstcorp.org/browse/IT-2613

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -118,7 +118,7 @@ mod 'lsst/ccs_mrtg',
   ref: 'v0.1.1'
 mod 'lsst/ccs_sal',
   git: 'https://github.com/lsst-it/puppet-ccs_sal.git',
-  ref: 'v0.2.1'
+  ref: 'v0.3.0'
 mod 'lsst/ccs_software',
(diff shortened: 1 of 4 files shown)


hreinking (approved):
LGTM, but, we will have the same conflict when rebasing @tony-johnson's branch.

tony-johnson:
One small comment, I also added:

> /etc/systemd/system/ocs-bridge-comcam-dev.service
> /etc/systemd/system/mcm-comcam-dev.service

both of which point to /lsst/ccs/dev rather than prod, and also in the case of the first one have an extra env var set. It would be good to get these included into the puppet setupm but it might need some discussion,

We will need to merge this into the IT2753 branch to deploy it on comcam.

jhoblitt:
I have no objection to this but it would be great if we could eventually merge this module with https://github.com/lsst-it/puppet-sal2 so we do not have to support two modules that do essentially the same thing.

glennmorris:
Although I believe it would be fine to merge this as-is, it may make sense to wait and include
https://github.com/lsst-it/puppet-ccs_sal/pull/4

PR #311, opened 2021-04-08
IT-2820: auxtel ccs

glennmorris:
These are the changes for management of CCS on auxtel hosts. They have been tested on the auxtel ccs hosts.
I don't believe there should be any effect on non-auxtel hosts.


glennmorris:
PS these changes require https://github.com/lsst-it/lsst-puppet-hiera-private/pull/44

hreinking (approved):
lgtm

PR #315, opened 2021-04-20
It 2753/comcam software

tony-johnson:
This pull request merges the comcam-software branch (IT-2753) back into master. Although the branch existed for some time there do not appear to be any current conflicts.

### DIFF
--- a/site/profile/manifests/ccs/postfix.pp
+++ b/site/profile/manifests/ccs/postfix.pp
@@ -8,6 +8,12 @@ class profile::ccs::postfix(
     ensure  => 'present',
     content => $auth,
   }
+  exec { 'sed -i \'/^::1/d\' /etc/hosts':
+    cwd      => '/tmp',
+    path     => ['/sbin', '/usr/sbin', '/bin'],
+    provider => shell,
+    onlyif   => 'grep \'::1         localhost\' /etc/hosts',
+  }
 
   ensure_packages($packages)
   Package[$packages] -> Class[postfix]
--- a/hieradata/cluster/comcam-ccs.yaml
+++ b/hieradata/cluster/comcam-ccs.yaml
@@ -112,4 +112,8 @@ ccs_software::installations:
     aliases:
       - "puppet-2.0.0"
 
+  comcam-software-2.0.1:
+    aliases:
+      - "puppet-2.0.1"
+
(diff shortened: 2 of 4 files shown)


glennmorris (approved):

site/profile/manifests/ccs/postfix.pp
```diff
@@ -8,6 +8,12 @@
     ensure  => 'present',
     content => $auth,
   }
+  exec { 'sed -i \'/^::1/d\' /etc/hosts':
```
jhoblitt, line 11: Puppet has a native {{host}} type for managing {{/etc/hosts}} entries.  See: https://puppet.com/docs/puppet/5.5/types/host.html  We are also already using a {{hosts}} (plural) module for managing {{/etc/hosts}} entries via hiera.  See https://github.com/lsst-it/lsst-itconf/blob/master/hieradata/node/atsdaq-mgmt.cp.lsst.org.yaml#L41-L43 for an example of doing it via hiera.  {{exec}} is generally a last resort operation.

jhoblitt (changes requested):

site/profile/manifests/ccs/postfix.pp
glennmorris: I don't know why that change was on this branch, maybe just expediency.
A simpler way to get the desired effect (which is "make postfix work") is perhaps just to add to cluster/comcam-ccs.yaml
`postfix::inet_protocols: "ipv4"`

tony-johnson: I do not know how that change got on to this branch, perhaps happened during rebasing confusion. @glennmorris Can you make your suggested change?

glennmorris: It's not my change. I just restored it after it got clobbered. (By the way, there may have been other clobbered changes that we are unaware of.) It came from here: https://jira.lsstcorp.org/browse/IT-2816
I could force push (...) a new version of this branch without it?

glennmorris: In the interest of moving things along, I created
https://github.com/lsst-it/lsst-itconf/pull/317
which is the same as this one, minus the /etc/hosts change.
So maybe you want to apply that one and close this one...

tony-johnson:
Closing in favor of #317

PR #317, opened 2021-04-21
It 2753/comcam software v2

glennmorris:


glennmorris:
This is intended to replace https://github.com/lsst-it/lsst-itconf/pull/315

PR #321, opened 2021-04-30
It 2935/pathfinder hcu01

tony-johnson:
Merge changes for pathfinder-hcu01. 

These add network settings for pathfinder-hcu01 and update the ccs versions to be installed.


glennmorris:
It seems to me that the non-backport parts of this pull request were applied some time ago, so I'm taking the liberty of closing it. Please reopen if I'm wrong.

PR #327, opened 2021-05-10
hiera: add rh-git218 to ccs clusters

glennmorris:
Installs the rh-git218 package from the centos-release-scl-rh repository


hreinking (approved):
lgtm

PR #349, opened 2021-06-28
IT-3024 update docker nat ignore rules for puppet firewall

andylytical:
Updates to prevent puppet firewall from removing NAT rules created by docker.


andylytical:
@jhoblitt FYI: tiny hiera update for puppet firewall at NCSA.

PR #403, opened 2021-07-30
Make puppet generate sudo files for ccs services, rather than relying on a cron job

glennmorris:
This is preparation for enabling purge of non-managed sudo files.


jhoblitt (approved):
This looks quite reasonable.  I'm very happy to see to be converging on a single method for managing local sudo rules.

PR #444, opened 2021-10-15
It 3204/comcam software

tony-johnson:
This is deployed on comcam and auxtel. and needs to be merged so it can be deployed on tucson test stand


jhoblitt (changes requested):
The diffs look good but could you do a rebase to remove the merge commits from the branch?

PR #474, opened 2021-11-12
Replace profile::ccs::clustershell with standard clustershell module

glennmorris:
Also, add site-specific configuration for tu ccs clusters.

(No rush at all with this.)


hieradata/site/tu/cluster/auxtel-ccs.yaml
```diff
@@ -6,3 +6,6 @@ ccs_monit::alert:
   ## tucson-teststand-alerts
   - "tucson-teststand-aler-aaaae4zsdubhmm3n7mowaugr2y@lsstc.slack.com"
 ccs_monit::hwraid: false
+
+clustershell::groupmembers:
+  all: {group: "all", member: "auxtel-fp01,auxtel-mcm"}
```
jhoblitt, line 11: :+1: 
I feel like we should be able to query either foreman or puppetdb for this information.  I spent maybe 10 mins looking through the clustershell docs looking for dynamic group members resolution and gave up.  We could do a `puppetdb_query()` from puppet code though.  Nothing urgent but it would be nice if we didn't have maintain static lists of hosts.

PR #480, opened 2021-11-30
Bump ccs_software module version

glennmorris:
This version produces alerts when a ccs systemd service fails.
Add the corresponding hiera config for the alert email address.


hreinking (approved):
lgtm

PR #488, opened 2021-12-09
DM-32881 bump NCSA default OS snapshot

andylytical:
Affects nodes

- lsst-nts-ccheader.ncsa.illinois.edu
- lsst-nts.csc1.ncsa.illinois.edu

Change are tested and applied to the nodes. This merge just get's rid of the branch and allows moving the `ncsa_production` pointer forward.


andylytical:
@jhoblitt - FYI to you. This should only affect the nodes mentioned above. All tests pass. I will merge to `master` shortly.

PR #582, opened 2022-06-23
(profile::core::common) rm ifcfg- file

jhoblitt:
Remove the ifcfg- file created on some hosts by problems with the
foreman provisioning template.

The presence of `/etc/sysconfig/network-scripts/ifcfg-` file on hosts running `libvirt` was causing strange errors via the API. Removing this fix fixed the errors.


cbarria (approved):
👍 this looks great, simple way to do what it was needed

PR #583, opened 2022-06-23
(hypervisor/dev role) update network config

jhoblitt:


cbarria (commented):
i get the config changes but i don't understand why are they needed, could you review the tittle of add comments please.

PR #585, opened 2022-06-24
update foreman to 3.2.1

jhoblitt:
Based on:

* #583
* #586
* #589

PR #586, opened 2022-06-24
(profile::core::puppet_master) fix cron /dev/null redirects

jhoblitt:
A typo (missing >) is causing error emails to pile up in
/var/spool/mail/root.


jhoblitt:
Split out of #585

cbarria (approved):

cbarria (approved):
The greater than symbol was needed to redirect the output the correct way

PR #587, opened 2022-06-24
(rke role) enable tab completion for k alias of kubectl

jhoblitt:


cbarria (approved):

cbarria (approved):
alias are always welcome 👍  also great for including test on something that is only use manually and not automated.

PR #589, opened 2022-06-29
(profile::core::debugutils) add jq package

jhoblitt:
Migrate jq package into profile::core::debugutils from the rke role to make it more widely available.


cbarria (approved):
moving JQ dependencies to a general role is great and incorporating tests for codes that are been worked on is always good! 👍

PR #590, opened 2022-06-30
hiera: no longer set ccs_monit::temp for comcam-fp01

glennmorris:
It just causes warnings.
Ref https://jira.lsstcorp.org/browse/IT-3814


jhoblitt (approved):
LGTM

PR #595, opened 2022-07-07
fix ipa client setup order of operations

jhoblitt:
The first puppet agent run on a newly provisioned hosts was failing due to auth configuration related resources attempting to sync prior to the `ipa-client-install` script having been run.


cbarria (commented):
references for the key were removed but the file added shouldn't be part of the same commit as it is not explained why.

jhoblitt:
> references for the key were removed but the file added shouldn't be part of the same commit as it is not explained why.

The commit is question was broken into two commits. The newly introduced spec file has been moved into its own commit.

cbarria (approved):
Changes on IPA config looks good (including test) but the Kerberos config arrange could be done on separately, up to you to leave them or put them on a different commit. 👍

PR #599, opened 2022-07-13
Switch to nexus as the source for CCS packages

glennmorris:
Fetch CCS packages from nexus instead of SLAC. This requires credentials from lsst-puppet-hiera-private.
Take the opportunity to use class parameters for the source URL (etc), rather the old global variable ccs_pkgarchive.


jhoblitt (approved):
:+1: for avoiding `lookup()` from puppet code.

PR #605, opened 2022-07-20
decouple profile mod spec from production hiera data

jhoblitt:
Unit tests of profile classes were, in some cases surprisingly, being effected by production hiera hierarchy data.  This is fixed by using a more limited fixtured hiera configuration by default and while the previous default fixtured hiera configuration is used only for example groups labeled with `:site`, Currently, only host/role specs are using `:site`.


cbarria (approved):
created custom configuration files with dummy parameters in order to successfully test without bringing default configuration that could cause this to fail. 👍

jhoblitt:
rebased on current master.

PR #606, opened 2022-07-25
Bump ccs java_artisanal version to 2.2.1

glennmorris:
This uses an explicit alternative for the java commands,
rather than relying on priority (which can be fragile).


hreinking (approved):
LGTM

PR #608, opened 2022-07-27
(profile::core::puppet_master) rm duplicate cron resource spec

jhoblitt:


cbarria (commented):
the webhook cron for this role is duplicated, but i dont see why the cron smee was removed

PR #610, opened 2022-07-28
add spec coverage for all roles

jhoblitt:
At least basic spec coverage for all roles is desired as preparation for adding EL8 support to most roles.  If an existing role does not have a dedicated `_spec.rb` file it will automatically be tested as if it is a "common" role.

based on:
- #609
- #611 
- #612 
- #613


spec/hosts/roles/ccs_mcm_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+role = 'generic'
```
jhoblitt, line 5: Wrong role name.

cbarria (approved):
Lots of changes from several commits, everything good 👍  
- removed file (ipa servers will use the ipareplica roll)
- dns forwarder for Tucson Site fixed
- rearrenged order hiera
- added test for most roles
- added selinux dependencie not declared 
- testing OS for ubuntu 20.04 and 22.04

PR #611, opened 2022-07-28
(dnscache role) rm site specific override of dns::forwarders

jhoblitt:
It is unknown why the tu site was not using 1.1.1.1 as a forwarder as
cp,dev,ls do.  Perhaps there was an access restriction on 1.1.1.1 in the
past?  There is no current issue with accessing 1.1.1.1 from
dns1.tu.lsst.org. E.g.:

    [jhoblitt@dns1 ~]$ dig +short google.com @1.0.0.1
    142.250.68.46
    [jhoblitt@dns1 ~]$ dig +short google.com @1.1.1.1
    142.250.72.142
    [jhoblitt@dns1 ~]$ dig +short google.com @8.8.8.8
    142.251.46.142


cbarria (approved):
tucson site had different config than other sites 👍

PR #613, opened 2022-07-28
(dm-bastion role) include profile::core::common

jhoblitt:
It is unknown as to why this role did not include authz management.


cbarria (approved):
added the common role for dm-bastion

PR #614, opened 2022-07-28
(common) fix setting of yum::plugin::versionlock::clean param

jhoblitt:
The current syntax is nonsense.  There is no `versionlock` param on the
`yum::plugin` class to pass a hash into.

based on:
-#610


cbarria (approved):
i saw these changes when reviewing #610 👍

PR #615, opened 2022-07-29
(Puppetfile) bump puppet/python to 6.3.0

jhoblitt:
No change in functionality. This is migrating from a git tag to an official release.


cbarria (approved):
move python to newer version 👍

PR #622, opened 2022-08-01
test classes/defines using facterdb data

jhoblitt:
Host/role specs were converted to use facterdb in #610 .


cbarria (commented):
only thing i see is different OS test where added (on_supported_os.each)

jhoblitt:
@cbarria This is changing from test with an enormous set of centos 7 test of facts (copied from one of our production hosts) in `spec/default_facts.yml` to using facterdb for most test data.  facterdb contains dumps of facts from all of the operatingsystems we are likely to use and saves us the effort of having to copy in test facts for EL7, EL8, etc.

cbarria:
great! thanks for the explanation, didn't knew what facterdb does, now i know why so much declarative code when away! 👍

jhoblitt:
`facterdb` is a ruby gem which was already present as a dep of `voxpupuli-test`.  https://github.com/voxpupuli/facterdb

cbarria (approved):
thanks for the explanation on facterdb 👍

PR #623, opened 2022-08-02
(spec) reduce role catchall to one site per OS at random

jhoblitt:
As a performance optimization, test with only one site.  If it is
important that a role is better tested, it should have its own spec
file.


cbarria (approved):
testing would be overkill if tried on every Site per OS 👍

jhoblitt:
I would prefer to test all sites but I couldn't figure out a sane way to parallelize the catchall tests.  If/when individual spec files are created for each role, then that role will be tested in parallel with other roles.

PR #634, opened 2022-08-10
Add FQDN entry to /etc/hosts on CCS clusters

glennmorris:
The FQDN entry should be there by default, but the upstream hosts module has a long-standing bug:
https://github.com/ghoneycutt/puppet-module-hosts/issues/40

This change switches to an lsst-it fork of the module, which adds this pull request:
https://github.com/ghoneycutt/puppet-module-hosts/pull/63

This fixes adding the FQDN entry, if stored_config = false (it is true by default).
It is set false only on CCS clusters (though I can't see why it should not be so everywhere).

Ref: https://jira.lsstcorp.org/browse/IT-3889

Note:
This uses the branch IT-3889/fqdn of lsst-it/puppet-module-hosts.
That probably isn't the right way to do it.
Maybe there should be an "lsst_production" (or somesuch) branch in that repo?


jhoblitt (approved):
This seems reasonable.  It does make me wonder if we need to add another hiera layer to allow data to be set for all subsystem related hosts.

PR #642, opened 2022-08-24
update to modern yum::versionlock api 

jhoblitt:
The semantics of the yum versionlock plugin changed between EL7 and EL8.  The `yum::versionlock` defined type added new parameters in order to be compatible with EL8+.  The new parameters are backwards compatible with EL7.


cbarria (approved):
update yum versionlock for el8 👍

PR #643, opened 2022-08-26
docker cleanup/update

jhoblitt:
Presently, all roles are using the same docker version.  Removing the hiera keys from the common hierarchy and relying the defaults is both DRYing and removes versionlocks which will have to be updated and maintained for EL8.


cbarria (approved):
part of the cleanup for el8 👍

PR #644, opened 2022-08-29
replace inline code with lsst/smee mod to gain EL8 support

jhoblitt:


cbarria (approved):
reference move to standalone 👍

PR #645, opened 2022-08-31
replace inline kubectl install code with lsst/kubectl mod

jhoblitt:
The existing `kubectl` method uses a Google provided yum repo and rpm to install kubectl.  Google does not provide an equivalent repo for EL8.  Although, the same binary/rpm should work on EL8, it would be ugly to configuring an EL7 yum repo across multiple EL major versions. As a more distribution agnostic approach, the `lsst/kubectl` puppet module downloads and installs official k8s `kubectl` binary releases.


cbarria (approved):
bad google for not upgrading those 👍

PR #646, opened 2022-08-31
fix ipareplica role versionlock release string

jhoblitt:


cbarria (approved):
fix IPAreplica role to fix version.

PR #648, opened 2022-08-31
rely on puppet/yum module to manage the default os yum repos for all members of the EL osfamily

jhoblitt:
This is required to support any member of the EL osfamily other than Centos 7.


cbarria (approved):
ver. bump and EL support 👍

PR #651, opened 2022-09-08
add filename extension to icinga-ipl archive

cbarria:
the file type is been checked by the extension of the file, not the content.


jhoblitt:
Closed in favor of #656

PR #652, opened 2022-09-09
allow ssh between saluser@tel-lt1.tu and saluser@tel-hw1.tu

jhoblitt:
Additionally, krb5 keytab values have been converted from `String` to `Sensitive[String]` as a first step towards more considerate handling of secrets.

See: https://puppet.com/docs/puppet/7/securing-sensitive-data.html

Related to https://github.com/lsst-it/lsst-puppet-hiera-private/pull/75

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -126,3 +126,5 @@ RSpec/ImplicitSubject:
   Enabled: false
 Style/CommentedKeyword:
   Enabled: false
+Naming/FileName:
+  Enabled: false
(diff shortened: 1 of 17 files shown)


cbarria (commented):
more secured data inside the configuration 👍

jhoblitt:
The puppet `Sensitive` variant type is used to wrap other types and requires an explicit `unwrap` call to stringify.  The intent to make it harder to accidentally print secrets into the logs.

cbarria (approved):
thanks for the explanation 👍

PR #655, opened 2022-09-13
(foreman role) set foreman_envsync image to v1.4.1

jhoblitt:


cbarria (approved):
goes with the version you fixed it 👍

PR #656, opened 2022-09-13
fix icinga director `Uncaught Error: Class 'Icinga\Util\Translator'` error

jhoblitt:
Additionally:

- Icinga related puppet modules are updated to the latest version to pickup AlmaLinux support.
- It is now possible to deploy the `icinga-master` role under the `dev` site.
- The `python-s3transfer` package is version pinned to fix `certbot`.
- Numerous fixes to allow the `icinga-master` role to be deployed onto a freshly provisioned host without existing state.
- Note that `pnp4nagios` (also called `pnp`) is effectively dead and we should migrate to influxdb/grafana.
  See: https://icinga.com/docs/icinga-2/latest/doc/13-addons/#influxdb


jhoblitt:
I've noticed that the icinga director daemon keeps being killed similar to https://github.com/Icinga/icingaweb2-module-director/issues/2577 It appears we may have to update php.

jhoblitt:
A php update does not appear to be required.  The mysterious `icinga-director` daemon restarting seems to be resolved.

cbarria (approved):
wans't a small fix, thanks👍

hreinking (approved):
Both Icinga instances seem to be working perfectly, nice work.
LGTM

PR #660, opened 2022-10-03
(rke role) ensure br_netfilter kmod is loaded

jhoblitt:
This mod provides the `net.bridge.bridge-nf-call-iptables` sysctl, which
rke requires be set to a value different than the default.  It seems
that this kmod often isn't loaded until dockerd starts up. This was
causing an order of operations issue with setting
the`net.bridge.bridge-nf-call-iptables` sysctl.


cbarria (approved):
kmod fix 👍

PR #664, opened 2022-10-06
IT-3996: dhcp for vlan1907

hreinking:


jhoblitt (changes requested):
Please add unit tests.

spec/hosts/nodes/foreman.cp.lsst.org_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      let(:node_params) do
+        {
+          role: 'foreman',
+          site: 'cp',
+          cluster: 'corecp',
+        }
+      end
+
+      let(:manifest) do
```
jhoblitt, line 22: In puppet, the `manifest` is the source code that is "compiled" into a `catalog`.  The goal of the role unit test(s) is to inspect the `catalog` that is created from compiling the profiles together with the hiera data.  The block of code in the `:manfiest` string here is adding `dhcp::pool` defined types into the manifest for this spec example but it doesn't help determine if the hiera data is resulting in the desired objects in the catalog.

What I suggest you do instead is inspect the catalog for `dhcp::pool` defined types. E.g.

(untested example)
```ruby
  it do
    is_expected.to contain_dhcp__pool('it_gss').with(
       network: '139.229.160.0',
       ...
    )
  end
```

jhoblitt: Also note that `foreman.cp` role is already being tested here: https://github.com/lsst-it/lsst-itconf/blob/master/spec/hosts/roles/foreman_spec.rb#L255-L269  The distinction between a "role" test a "node" test is that "node" is covering things they are truly specific to a single node. IOW - the data from the `hierdata/node` layer.

hreinking: Sounds perfect! I'll apply the changes. Regarding the role test vs node test, do you want me to add and exclude foreman/cp from the role test?

hieradata/site/cp/role/foreman.yaml
```diff
@@ -64,27 +40,28 @@ dhcp::pools:
     network: "139.229.164.0"
     mask: "255.255.255.0"
     gateway: "139.229.164.254"
-    range: "139.229.164.1 139.229.164.200"
+    range:
+      - "139.229.164.1 139.229.164.200"
     search_domains: "%{alias('dhcp::dnsdomain')}"
-  cds-arch:
+  cds_arch:
```
jhoblitt, line 46: The resource names in puppet and added to the dhcp.conf as a comment.  I would like this name to be match the subnet name in ipam and in foreman.  The foreman subnet names are largely inconsistent but I think the values in ipam are mostly in sync.  E.g. https://ipam.ls.lsst.org/index.php?page=subnets&section=1&subnetId=16

I don't understand the motivation for converting `-`'s into `_` here but if we are going to do it, the subnets name in foreman and ipam should be sync'd at the same time.

jhoblitt: @julioconstanzo Do you have an opinion about changing the vlan names in the switches use `_` instead of `-`?

hreinking: regarding de - to _, just standardization; I'll update names so they match IPAM.

jhoblitt (approved):
LGTM.

hieradata/site/cp/role/foreman.yaml
jhoblitt: It looks like the transliteration to underscores has been backed out, so I'll mark this conversation as resolved.

PR #668, opened 2022-10-13
manage auditd rules

jhoblitt:
The `simp/auditd` mod includes a default set of auditd rules which appear to be a reasonable starting point.


jhoblitt:
Should be good to go now.

cbarria (approved):
great comments on default_facts 👍

PR #681, opened 2022-11-08
add basic hiera config for lsstcam-ccs cluster

glennmorris:


jhoblitt (approved):
Seems reasonable to me. We need to resolve some of these FIXMEs in the near future.

PR #682, opened 2022-11-08
hiera: update the source URL for CCSWebTrending.war

glennmorris:
To something that actually exists, ref eg
https://lsstc.slack.com/archives/C01270EHCQ3/p1667925293321179


jhoblitt (approved):
LGTM. I've rebased on current master.

PR #683, opened 2022-11-09
(docker-compose role) use docker-compose package

jhoblitt:
Instead of manual installation of `docker-compose` using pip.

To resolve this error:

    Error: 'pip3 --log /tmp/pip.log install     docker-compose==1.25.0' returned 1 instead of one of [0]
    Error: /Stage[main]/Python/Python::Pip[docker-compose]/Exec[pip_install_docker-compose]/returns: change from 'notrun' to ['0'] failed: 'pip3 --log /tmp/pip.log install     docker-compose==1.25.0' returned 1 instead of one of [0] (corrective)

### DIFF
--- a/hieradata/role/amor.yaml
+++ b/hieradata/role/amor.yaml
@@ -3,24 +3,8 @@ classes:
   - "profile::core::common"
   - "profile::core::debugutils"
   - "profile::core::docker"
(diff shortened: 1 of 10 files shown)


cbarria (approved):
Looks Great 👍

PR #694, opened 2022-11-16
tweak chronyd config

jhoblitt:
- set logchange value to 5ms (1/2 the max clock error of 10ms per ITTN-009)
- disable chronyd command port; chronyc will work via the unix domain socket when run as root
- explicitly disable active as as ntp server
- remove the local stratum config as it is a noop in client only mode


cbarria (approved):
Looks Great 👍

PR #698, opened 2022-11-21
(gha) switch to 16 core runner

jhoblitt:
The gha run time is already > 10 minutes with only two OS in the test matrix (EL7 & EL8) when using the default gha runner.  It is hoped that switching to the 16c runner will reduce the time for all gha workflows to run to < 10 minutes.


cbarria (approved):
this is great! thanks 👍

PR #703, opened 2022-11-21
bump docker to 20.10.21

jhoblitt:
The currently pinned version, 20.10.12, does not have packages for EL9.
We are updating to a version that has packages for EL7, EL8, & EL9.

Based on #702


cbarria (approved):
looks great 👍

PR #707, opened 2022-11-21
disable icinga on EL9

jhoblitt:
There are no freely available official packages built/published for EL9 as icinga has decided to place them behind a paywall.  We need to consider if we want to build our own rpms from source, license icinga, or migrate to a different solution. In the meantime, we need to disable attempting to manage icinga on EL9 via puppet.

See: https://github.com/Icinga/icinga2/issues/9390

Based on #704


cbarria (approved):
Looks Great 👍

PR #708, opened 2022-11-21
(profile::core::debugutils) drop tuned-utils-systemtap package

jhoblitt:
As this package isn't present on EL9 and it is unclear if we have ever used this package.


cbarria (approved):
Looks Great 👍

PR #710, opened 2022-11-21
(rke role) restrict inclusion of clustershell class to EL7/EL8

jhoblitt:
There does not [yet] appear to be a clustershell package in EPEL9.


cbarria (approved):
Looks Great 👍

PR #712, opened 2022-11-21
(profile::core::rke) drop class level require of easy_ipa

jhoblitt:
The class level require was effectively blocking the docker class resources if easy_ipa fails to install.  Instead of the entire class requiring easy_ipa, we are adding the require metaparam to two resources that need ipa user name resolution to be functioning.


cbarria (approved):
Looks Great 👍

PR #719, opened 2022-12-22
add EL9 Network Config Manke Cluster

cbarria:
adds the basic initial configuration for network interfaces, this could probably be moved to another hierarchy.


hreinking:
Please add the test modules.

PR #722, opened 2023-01-11
(profile::core::common) enable NetworkManager support on EL9 by default

jhoblitt:
This change will enable NetworkManager support on EL9 nodes by default. The current scheme requires each role to "opt-in" to NM support.


cbarria (approved):
Looks Great 👍🏻

PR #724, opened 2023-01-20
Updating DAQ release to R5-V5.0

jgthayer:


jgthayer:
This has been installed on the TTS and appears to be working. I think we're ready to pull, change the puppet environment and close the IT Jira. I can do as many of those things as you like.

jhoblitt (approved):
LGTM.

PR #727, opened 2023-01-27
(foreman.ls) add BTS pools

cbarria:
this add 4 Pools to the BTS Network
- LSSTCAM hosts
- AUXTEL hosts
- MISC hosts (host none dss related)
- LHN hosts (for host that need a second interface to LHN)


hieradata/site/ls/role/foreman.yaml
```diff
@@ -92,12 +92,36 @@ dhcp::pools:
...
     mask: "255.255.255.0"
     gateway: "139.229.151.254"
     range:
       - "139.229.151.201 139.229.151.249"  # ~ /27
+  BTS_LSSTCAM:
+    network: "139.229.152.0"
+    mask: "255.255.255.128"
+    gateway: "139.229.152.126"
+    range:
+      - "139.229.152.70 139.229.152.120"  #
```
jhoblitt, line 106: The dangling comment is odd. It would be good to either remove it or add the prefix length to be consistent with the other dhcp pools.

jhoblitt (changes requested):
One small complaint about comments but otherwise looks good.

hieradata/site/ls/role/foreman.yaml
cbarria: you are right! i used an old line and copy/paste it... i will remove it

jhoblitt (approved):

PR #739, opened 2023-02-09
(proflie::core::sysctl::rp_filter) add 'all' & 'default' to dev list

jhoblitt:
Testing has shown that setting `net.ipv4.conf.p2p1/2505.rp_filter` to
`0` has no effect if `net.ipv4.conf.all.rp_filter` is set to `1`. Thus,
we need to also change the value of `net.ipv4.conf.all.rp_filter` when
modifying the other interfaces. For good measure,
`net.ipv4.conf.default.rp_filter` is also set to match
`net.ipv4.conf.all.rp_filter` to handle any newly created interfaces.


cbarria (approved):
Looks Great 👍🏻

PR #741, opened 2023-02-09
IT-4253: GIS bastion

hreinking:


spec/hosts/nodes/bastion1.cp.lsst.org_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+#
+# note that this hosts has interfaces with an mtu of 9000
+#
+describe 'bastion1.cp.lsst.org', :site do
```
jhoblitt, line 8: There is no `hieradata/node/bastion1.cp.lsst.org.yaml` file in the repo. Is this intended to be a role test? If so, it should under `spec/hosts/roles/`

hreinking: yeap, you're right, not needed

jhoblitt (approved):
LGTM.

PR #742, opened 2023-02-09
(foreman.ls) add SLAC routes to LHN network and tests

cbarria:
Separated this branch from the test because this routes are needed either way (test and production) 
We are adding just the prefix from 134.79.XXX.XXX from the border router to the LHN DHCP Network routes

`ip prefix-list LHN-BGP-In: 
   seq 70 permit 134.79.20.0/23 
   seq 75 permit 134.79.235.224/28 
   seq 80 permit 134.79.235.240/28 
   seq 85 permit 134.79.23.0/24`


jhoblitt (approved):
LGTM

PR #750, opened 2023-03-01
(site/ls/role/foreman) rm 139.229.152/25 gw (dhcp routers option)

jhoblitt:
This was causing multus pods to have a double default gw.  Attempts to
change the NAD to disable setting a default gw failed.  The options are
to either hand out bogus static routes as part of option 121 or to not
send the routers option.


cbarria (approved):
Looks Great 👍🏻

PR #754, opened 2023-03-03
(role/foreman) retire {dev,tu,ls}_production env name

jhoblitt:
The `{dev,tu,ls}_production` envs are to be replaced with `production`.
The `cp_production` env is still retained for use at the summit.

Based on #753


cbarria (approved):
Looks Great 👍🏻

PR #759, opened 2023-03-08
(dimm.cp) add network config and tests

cbarria:
the Dimm was migrated from a VM to a Bare Metal fanless server and installed with Alma 8, there was no networking config and puppet was deleting the interfaces on each run, so we created the network configuration for it.

### DIFF
--- a/hieradata/node/dimm.cp.lsst.org.yaml
+++ b/hieradata/node/dimm.cp.lsst.org.yaml
@@ -1,0 +1,33 @@
+profile::nm::connections:
+  eno1:
+    content: |
+      [connection]
+      id=eno1
+      uuid=250c1a0b-95c3-4798-a4df-d864e60d3b63
+      type=ethernet
+      autoconnect=true
+      interface-name=eno1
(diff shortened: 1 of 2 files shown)


hieradata/node/dimm.cp.lsst.org.yaml
```diff
@@ -0,0 +1,33 @@
...
+      id=enp2s0
+      uuid=de9904c8-9577-1a17-36b1-34b94132f06a
+      type=ethernet
+      autoconnect=true
+      interface-name=enp2s0
+
+      [ethernet]
+
+      [ipv4]
+      method=auto
```
hreinking: Isn't the secondary NIC use for a point to point connection? Maybe ask Brian which IP it needs and leave it static

cbarria: very true sir, lets use this opportunity to leave it static with the correct ip/net, thanks

hreinking (approved):
lgtm

PR #760, opened 2023-03-09
(ccs) add hiera settings for BTS

glennmorris:
I'm assuming that the names for the dds interfaces will be auxtel-mcm-dds.ls.lsst.org and lsstcam-mcm-dds.ls.lsst.org,
consistent with eg the summit and Tucson, but the names don't seem to exist yet?

### DIFF
--- a/hieradata/site/ls/cluster/auxtel-ccs.yaml
+++ b/hieradata/site/ls/cluster/auxtel-ccs.yaml
@@ -1,0 +1,15 @@
+---
+ccs_site: "base"
+ccs_sal::dds_interface: "auxtel-mcm-dds.ls.lsst.org"
+
+## base-teststand-alerts
+ccs_software::service_email: "base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com"
+
+ccs_monit::alert:
+  - "base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com"
+
+clustershell::groupmembers:
+  misc: {group: "misc", member: "auxtel-mcm,auxtel-dc01,auxtel-fp01"}
+  #hcu: {group: "hcu", member: "auxtel-hcu01"}
+  #all: {group: "all", member: "@misc,@hcu"}
+  all: {group: "all", member: "@misc"}
--- a/spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
+++ b/spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
@@ -21,6 +21,7 @@ describe 'lsstcam-db01.ls.lsst.org', :site do
         {
           role: 'ccs-database',
           site: 'ls',
+          cluster: 'lsstcam-ccs',
         }
       end
 
--- a/spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
+++ b/spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
@@ -24,6 +24,10 @@ describe 'lsstcam-db01.ls.lsst.org', :site do
         }
       end
 
+      it { is_expected.to contain_file('/etc/ccs/systemd-email').with_content(%r{^EMAIL=base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com}) }
+
+      it { is_expected.to contain_file('/etc/monit.d/alert').with_content(%r{^set alert base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com}) }
+
       it { is_expected.to compile.with_all_deps }
 
       include_context 'with nm interface'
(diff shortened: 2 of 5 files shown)


glennmorris:
I took the opportunity to add a related second commit.

cbarria (commented):
i don't see any unit testing on these cluster settings, probably would be good idea to add them on the nodes that already have test. 
- lsstcam-dc01.ls.lsst.org_spec.rb
- auxtel-dc01.ls.lsst.org_spec.rb

glennmorris:
> i don't see any unit testing on these cluster settings, probably would be good idea to add them on the nodes that already have test.
> 
> * lsstcam-dc01.ls.lsst.org_spec.rb
> * auxtel-dc01.ls.lsst.org_spec.rb

I added some in 6268a30, but I am not good with tests.
Can anyone tell me why the settings for ccs_monit::alert and ccs_software::service_email seemingly aren't being applied?

If it is because the tests run in a sparse environment without the real per-host hiera settings, then I don't understand what I am being asked to test, since this change is nothing but hiera settings.

glennmorris:
Furthermore, I think the tests I added don't belong here, but rather in the puppet-ccs_software and puppet-ccs_monit modules. So I am confused about what tests are wanted for this pull request.

hieradata/site/ls/cluster/auxtel-ccs.yaml
```diff
@@ -0,0 +1,15 @@
...
+ccs_software::service_email: "base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com"
+
+ccs_monit::alert:
+  - "base-teststand-alerts-aaaai5j4osevcaaobtog67nxlq@lsstc.slack.com"
+
+clustershell::groupmembers:
+  misc: {group: "misc", member: "auxtel-mcm,auxtel-dc01,auxtel-fp01"}
+  #hcu: {group: "hcu", member: "auxtel-hcu01"}
+  #all: {group: "all", member: "@misc,@hcu"}
+  all: {group: "all", member: "@misc"}
```
hreinking, line 15: I believe @cbarria is referring for tests of these values (also for the ones in lsstcam-ccs), which are very similar as the ones he added on 
https://github.com/lsst-it/lsst-control/pull/762. 
For the clustershell::groupmembers, you can find one as example at 
https://github.com/lsst-it/lsst-control/blob/production/spec/hosts/nodes/manke01.ls.lsst.org_spec.rb#L30-L39
btw, I would add the test for auxtel-ccs at auxtel-dc01.ls.lsst.org_spec.rb
https://github.com/lsst-it/lsst-control/blob/production/spec/hosts/nodes/auxtel-dc01.ls.lsst.org_spec.rb 
and for lsstcam-ccs at lsstcam-dc01.ls.lsst.org_spec.rb
https://github.com/lsst-it/lsst-control/blob/production/spec/hosts/nodes/lsstcam-dc01.ls.lsst.org_spec.rb

glennmorris: I did try to add some tests like those, in https://github.com/lsst-it/lsst-control/pull/760/commits/6268a30482b63d830949df14e2f8caa5039a8994 
but as I said, 2/3 of them fail because the hiera settings for ccs_software::service_email and ccs_monit::alert: are apparently not being set in the test environment. Can someone explain to me how it works?

I can try to add tests for the clustershell setting too, but then surely I would just be testing that the clustershell puppet module works, which is something that should be done in that module's own tests?

glennmorris: Oh, maybe now I see: lsstcam-db01.ls.lsst.org_spec.rb has no cluster setting...

glennmorris: Adding the cluster setting revealed that the network tests weren't correct.

glennmorris: Tests added, please re-review.

spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
```diff
@@ -52,6 +57,32 @@
         it { expect(nm_keyfile['connection']['type']).to eq('ethernet') }
         it { expect(nm_keyfile['connection']['autoconnect']).to be_nil }
       end
+
+      context 'with enp129s0f1.2505' do
```
jhoblitt, line 61: Does -db01 need an LHN interface?

glennmorris: I don't know what's going on here, I just made the test consistent with the existing hiera.
I note that profile::nm::connections in node/lsstcam-db01.ls.lsst.org.yaml is identical to that in site/ls/cluster/lsstcam-ccs.yaml, but with enp129s0f1.2505 and br2505 removed.
On the face of it, this does not make sense, given the merge strategy for this item.

glennmorris: Maybe there needs to be a separate clean-up of this area?

jhoblitt: The live system already has an lhn interface, so it doesn't make sense to hold on merging this PR, but we should probably fix that.

jhoblitt (approved):

spec/hosts/nodes/lsstcam-db01.ls.lsst.org_spec.rb
jhoblitt: Jira issue to remove the lhn interfaces from hosts which should not have it: https://jira.lsstcorp.org/browse/IT-4387

PR #765, opened 2023-03-10
(konkong) initial network config and test

cbarria:
This adds the initial network configuration and the dds configuration (no LHN) also, has tests.

### DIFF
--- a/hieradata/cluster/konkong.yaml
+++ b/hieradata/cluster/konkong.yaml
@@ -1,0 +1,131 @@
+---
+clustershell::groupmembers:
+  konkong: {group: "konkong", member: "konkong[01-03]"}
+cni::plugins::checksum: "f3a841324845ca6bf0d4091b4fc7f97e18a623172158b72fc3fdcdb9d42d2d37"
+cni::plugins::enable: ["macvlan"]
+cni::plugins::version: "1.2.0"
+profile::core::common::manage_powertop: true
+profile::core::ospl::enable_rundir: true
+profile::core::rke::enable_dhcp: true
+profile::core::rke::version: "1.3.12"
+profile::nm::connections:
+  eno1np0:
+    content: |
+      [connection]
+      id=eno1np0
+      uuid=f330f829-20cc-b829-67b0-18086a5fe6fa
+      type=ethernet
+      autoconnect=false
+      interface-name=eno1np0
+
+      [ethernet]
+
+      [ipv4]
+      method=disabled
+
+      [ipv6]
+      method=disabled
+  eno2np1:
+    content: |
+      [connection]
+      id=eno2np1
+      uuid=de9904c8-9577-1a17-36b1-34b94132f06a
+      type=ethernet
+      autoconnect=false
+      interface-name=eno2np1
+
+      [ethernet]
+
+      [ipv4]
+      method=disabled
+
+      [ipv6]
+      method=disabled
+  enp4s0f3u2u2c2:
+    content: |
+      [connection]
+      id=enp4s0f3u2u2c2
+      uuid=283f3035-13d7-4c87-9d7a-7d47861fa1f9
+      type=ethernet
+      autoconnect=false
+      interface-name=enp4s0f3u2u2c2
+
+      [ethernet]
+
+      [ipv4]
+      method=disabled
+
+      [ipv6]
+      method=disabled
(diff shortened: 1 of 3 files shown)


hieradata/cluster/konkong.yaml
```diff
@@ -0,0 +1,126 @@
...
+      [ethernet]
+
+      [bridge]
+      stp=false
+
+      [ipv4]
+      method=disabled
+
+      [ipv6]
+      method=disabled
```
hreinking: I would suggest to move the config into site/ls/cluster/konkong.yaml

hieradata/cluster/konkong.yaml
```diff
@@ -0,0 +1,126 @@
...
+      interface-name=enp129s0f1
+
+      [ethernet]
+
+      [ipv4]
+      method=disabled
+
+      [ipv6]
+      method=disabled
+  enp129s0f1.2502:
```
hreinking: Don't know if it's typo, but this is DDS network, LHN one is 2505

hieradata/cluster/konkong.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      [ethernet]
+
+      [vlan]
+      flags=1
+      id=2502
+      parent=enp129s0f1
+
+      [bridge-port]
+  br2502:
```
hreinking: If no DDS network is intended, for consistency, I would change it to br2505

hreinking (changes requested):
Please review the comments.

hieradata/cluster/konkong.yaml
cbarria: i just use the same hierachy as Manke, if this is wrong, we should change them both. but i think its OK, all the other clusters have the same config here.
imagen

hieradata/cluster/konkong.yaml
cbarria: you are right, interfaces should be for vlan 2505, not the DDS, will change the config, thanks

hieradata/cluster/konkong.yaml
cbarria: yes, the same as last comment, thanks.

hieradata/cluster/konkong.yaml
hreinking: yeap, you're right; highly unlikely we'll have another cluster with the same name

spec/hosts/nodes/konkong01.lsst.org_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      it do
+        is_expected.to contain_class('profile::core::sysctl::rp_filter').with_enable(false)
+      end
+
+      it do
+        is_expected.to contain_class('clustershell').with(
+          groupmembers: {
+            'konkong' => {
+              'group' => 'konkong',
+              'member' => 'konkong[01-03]',
```
hreinking, line 35: I would suggest do from 1-5 now, do a short branch just for that would be more pain in the future...

hieradata/cluster/konkong.yaml
```diff
@@ -0,0 +1,131 @@
+---
+clustershell::groupmembers:
+  konkong: {group: "konkong", member: "konkong[01-03]"}
```
hreinking, line 3: Same as above, just suggesting to go 1-5

hreinking (changes requested):
Just suggestions.

spec/hosts/nodes/konkong01.lsst.org_spec.rb
cbarria: i wont be able to avoid making little changes once the two nodes get included, not only here, but in the rook-ceph folder and the rke also, so i will have to do another branch to "activate" those 2 more nodes anyway.

hieradata/cluster/konkong.yaml
cbarria: same comment as above, changes will have to be made when added either way

jhoblitt: It is better that the clustershell config only references nodes which exist. Otherwise, clush commands will always return an error for those nodes.

jhoblitt (approved):

hreinking (approved):

PR #770, opened 2023-03-20
(role/ccs-dc) add ccs_daq module

glennmorris:
Add the ccs_daq module to the ccs-dc role.
Ref [ticket](https://jira.lsstcorp.org/browse/IT-4396).


jhoblitt (approved):
This looks good. I've added a commit to update the `ccs-dc` role spec. I'll merge when CI passes.

PR #771, opened 2023-03-20
(role/foreman) fix foreman_envsync tag name

jhoblitt:
The revised gha workflow is not creating the tag as `X.Y.Z` instead of `vX.Y.Z`.


cbarria (approved):
Looks Great 👍🏻

PR #773, opened 2023-03-21
(node/lsstcam-db01) run ccs services localdb and rest-server

glennmorris:
I thought about doing this in role/ccs-database instead, but it seems ccs_software::services has no merge strategy set at present, so it would have been a more invasive change.

### DIFF
--- a/hieradata/node/lsstcam-db01.cp.lsst.org.yaml
+++ b/hieradata/node/lsstcam-db01.cp.lsst.org.yaml
@@ -1,4 +1,9 @@
 ---
+ccs_software::services:
+  prod:
+    - "rest-server"
+    - "localdb"
+
 network::interfaces_hash:
   em1:  # fqdn
     bootproto: "dhcp"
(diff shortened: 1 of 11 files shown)


hreinking (approved):
lgtm

cbarria (changes requested):
changes look good, but there are no test for the services added... they should be somewhere around:

`it { is_expected.to contain_class('Ccs_software::Service')}
  it { is_expected.to contain_service('rest-server')}
 it { is_expected.to contain_service('localdb')}`
 
 👍🏻

jhoblitt:
As we haven't been requesting spec coverage of the `ccs_software` class parameters in the past, and there are no existing examples in the tree, I've added some commits to this branch which add this coverage.  I have also rebased on current `production`.

cbarria (approved):
Looks Great 👍🏻

jhoblitt (approved):

PR #777, opened 2023-03-30
(kueyen) Add support config for EL8

cbarria:
All the configuration for EL7 was removed and exchanged for network configuration compatible with EL8/9, plus some extra config for rke/clusters roles.


hreinking (approved):
lgtm

PR #778, opened 2023-04-04
(chonchon) add LHN network config vlan1800

cbarria:
needed configuration to be able to access the LHN from k8s.

### DIFF
--- a/hieradata/cluster/chonchon.yaml
+++ b/hieradata/cluster/chonchon.yaml
@@ -1,3 +1,44 @@
 ---
 clustershell::groupmembers:
   chonchon: {group: "chonchon", member: "chonchon[01-03]"}
+cni::plugins::checksum: "f3a841324845ca6bf0d4091b4fc7f97e18a623172158b72fc3fdcdb9d42d2d37"
+cni::plugins::enable: ["macvlan"]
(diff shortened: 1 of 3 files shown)


hieradata/cluster/chonchon.yaml
```diff
@@ -1,3 +1,45 @@
 ---
 clustershell::groupmembers:
   chonchon: {group: "chonchon", member: "chonchon[01-03]"}
+cni::plugins::checksum: "f3a841324845ca6bf0d4091b4fc7f97e18a623172158b72fc3fdcdb9d42d2d37"
+cni::plugins::enable: ["macvlan"]
+cni::plugins::version: "1.2.0"
+profile::core::common::manage_powertop: true
+profile::core::rke::enable_dhcp: true
+profile::core::ospl::enable_rundir: true
```
jhoblitt: This isn't needed as opensplice won't be running on this cluster.

jhoblitt (changes requested):

hieradata/cluster/chonchon.yaml
cbarria: thanks! was removed from the config.

jhoblitt (approved):
LGTM

PR #784, opened 2023-04-12
(dimm.cp) add NFS mount and telnet

cbarria:
This adds the telnet package to the Dimm, but also sets the NFS mount requested in IT-4421.


jhoblitt (approved):
LGTM.

PR #789, opened 2023-04-14
(ccs) small fixes for auxtel services

glennmorris:
This is the first 2 items from https://jira.lsstcorp.org/browse/IT-4438 .
The third item will require a bit more work, and will be done later.


jhoblitt (approved):
LGTM.

PR #790, opened 2023-04-14
(site/ls) refactor and dedup NetworkManger configuration

jhoblitt:
Summary of changes:

* Deduplicate NM keyfile configs which were cut'n'pasted between hosts.
* Consistently create bridge interfaces named `lsst-daq`, `dds`, or `lhn` on end-user facing hosts; remove `lhn` interfaces from hosts were it was inappropriate.
* Normalize keyfile config between hosts -- E.g. an interface configured for dhcp should have the same keyfile sections defined on every host.
* Fix multiple default routes being set from dhcp on multi-homed hosts.
* Improve unit testing of NM keyfiles.


cbarria (approved):
Looks Great 👍🏻

PR #791, opened 2023-04-17
(ccs) adjust settings of the ccs -ih service

glennmorris:
Specify a different User, Group, WorkingDirectory, and PATH environment for this service.


jhoblitt (approved):
This is really begging for deduplication.  Could we merge the `atsdaq` and `comcam-fp` then move the services up to the role level?

PR #793, opened 2023-04-18
(role/perfsonar) fix version pinning by controlling yumrepo url

jhoblitt:


cbarria (approved):
Looks Great! 👍🏻  Thanks for the fix

PR #794, opened 2023-04-18
(profile::core::perfsonar) rm perfSONAR yumrepo mirrorlist

jhoblitt:
The yumrepo type, by default, will not remove unmanaged parameters. The mirrorlist parameter must be explicitly set to 'absent' to remove it.


cbarria (approved):
Looks Great 👍🏻  thanks!

PR #799, opened 2023-04-20
(site/ls/role/daq-mgt) daq v6.1

jgthayer:
R5-V6.1 has been loaded on the BTS and appears to be working. This change was based on IT-4373 ~~, so that's where we can merge this back to.~~


jhoblitt (approved):
I'm slightly surprised this didn't require a unit test change but looks fine.

PR #800, opened 2023-04-20
(profile::daq::daq_interface) drop EL8+ support

jhoblitt:
This reverts the EL8 support from the `profile::daq::daq_interface` class as we have already decided not to continue to use this approach for configuring daq interfaces under EL8.

Split out of #790


glennmorris (approved):

cbarria (approved):
Looks Great! 👍🏻

PR #801, opened 2023-04-21
(node/azar03.cp) update network config

cbarria:
Network configuration has been updates due to hardware change on the host


jhoblitt (changes requested):
Please change the commit message to reference the heira layer. E.g. (node/azar03.cp[.lsst.org]). Otherwise, looks good.

jhoblitt (approved):

PR #802, opened 2023-04-21
(node/perfsonar.cp) add network config and routes

cbarria:
New perfsonar on the summit needs networking and routing configuration.


jhoblitt (changes requested):
Please update the commit message to reference the hiera layer. Otherwise, this looks good to go.

jhoblitt (approved):

PR #804, opened 2023-04-25
(Gemfile) pin puppet ~> 7.24

jhoblitt:
Puppet 8.0.0 has been released and is not compatible with Ruby < 3.2.


cbarria (approved):
Looks Great 👍🏻

PR #805, opened 2023-04-25
(role/dimm) fix file/dir ownership; fix serial device ownership

jhoblitt:
Fixes two issues observed on `dimm.cp.lsst.org`:

* The assorted dimm related services are running as `dimm:users`. This prevented them being able to create log files under the path `/mnt/dimm`, as those dirs were created by the rpm with a different `<uid>:<gid>`.
* Several of the dimm daemons are also trying to open `/dev/ttyUSB[012]`. By default, serial device ttys are owned `root:dialout`. As none of the running services where in the `dialout` group, they were unable to open these files.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -69,6 +69,9 @@ lookup_options:
     merge:
       strategy: "deep"
       knockout_prefix: "--"
+  systemd::udev_rules:
+    merge:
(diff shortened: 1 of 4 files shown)


cbarria (approved):
Looks Great, thanks!

PR #807, opened 2023-04-26
(role/daq-mgt) Update AuxTel to DAQ R5-V6.1

jgthayer:
The update has been installed and it looks like everything is working.


jhoblitt (approved):
LGTM.

PR #810, opened 2023-05-01
(ccs) bump DAQ version to R5-V6.1

glennmorris:
Once @tony-johnson approves I will request wider review.

### DIFF
--- a/hieradata/role/atsdaq.yaml
+++ b/hieradata/role/atsdaq.yaml
@@ -1,6 +1,7 @@
 # XXX rename to auxtel-fp
 ---
 classes:
+  - "ccs_daq"
   - "profile::ccs::common"
   - "profile::ccs::krb5_token"
   - "profile::core::common"
--- a/hieradata/site/cp/cluster/auxtel-ccs.yaml
+++ b/hieradata/site/cp/cluster/auxtel-ccs.yaml
@@ -1,2 +1,2 @@
 ---
-daq::daqsdk::version: "R5-V5.0"
+daq::daqsdk::version: "R5-V6.1"
--- a/hieradata/site/cp/cluster/comcam-ccs.yaml
+++ b/hieradata/site/cp/cluster/comcam-ccs.yaml
@@ -1,2 +1,2 @@
 ---
-daq::daqsdk::version: "R5-V5.0"
+daq::daqsdk::version: "R5-V6.1"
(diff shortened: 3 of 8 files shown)


tony-johnson (commented):
Looks good to me. 

Do we know why this does not appear to have been applied in the past on the auxtel summit machine? Or is auxtel-summit somehow configured to put the install in a different place?

tony-johnson (approved):

glennmorris:
> Do we know why this does not appear to have been applied in the past on the auxtel summit machine?

I think either it was done by hand, or by some older puppet method and then got accidentally left out.

hieradata/role/atsdaq.yaml
```diff
@@ -1,6 +1,7 @@
 # XXX rename to auxtel-fp
 ---
 classes:
+  - "ccs_daq"
```
jhoblitt, line 4: I think that `hieradata/site/tu/role/atsdaq.yaml` is redundant after this change:

```
$ cat ./site/tu/role/atsdaq.yaml
---
classes:
  # XXX this should migrate up to the top level role but it is first being
  # tested in the TTS
  - "ccs_daq"
```

I'm also wondering if we shouldn't delete the `atsdaq` role completely and switch those nodes to `comcam-fp` as they will be the essentially the same after this change.  Note that `dhcp::disable` can be removed from `comcam-fp` at this point... that is wrong when the daq submit dhcp server was migrated to daq-mgt* nodes.

```
$ diff atsdaq.yaml comcam-fp.yaml
--- atsdaq.yaml	2023-05-02 11:24:13.696118966 -0700
+++ comcam-fp.yaml	2023-04-21 08:39:18.650598980 -0700
@@ -1,8 +1,9 @@
-# XXX rename to auxtel-fp
 ---
 classes:
   - "ccs_daq"
+  - "dhcp::disable"
   - "profile::ccs::common"
+  - "profile::ccs::graphical"
   - "profile::ccs::krb5_token"
   - "profile::core::common"
   - "profile::core::debugutils"
```

glennmorris: Thanks for spotting this, I have added removal of site/tu/role/atsdaq to this pull.

I suggest merging the comcam and auxtel roles be a separate pull. Maybe also rename to ccs-fp?

jhoblitt: > I suggest merging the comcam and auxtel roles be a separate pull. Maybe also rename to ccs-fp?
:+1:

jhoblitt (approved):
looks good now.

PR #811, opened 2023-05-02
(tu/cluster/auxtel-ccs) use ccs lockmanager

glennmorris:


hieradata/node/auxtel-mcm.tu.lsst.org.yaml
```diff
@@ -47,3 +47,7 @@ nfs::client_mounts:
     share: "ccs-data"
     server: "auxtel-fp01.tu.lsst.org"
     atboot: true
+
+ccs_software::services:
```
jhoblitt, line 51: Is the intent to role this out first for tts and then to bts & summit?  If so, I'm fine with merging this as-is but I'm wondering if it might make more sense to put it on the atsccs role (by site) rather than on the host?

glennmorris: Yes, the intent for now is just to have this on tts, and then at some later date on other hosts.

jhoblitt (approved):
Merging after discussion via slack.

PR #812, opened 2023-05-03
(cluster/ruka) rke version bump 1.4.6-rc4

cbarria:
This add support to update the current rke version to 1.4.5 so we can move k8s version to 1.25.9 on ruka atm.


dtapiacl (commented):
Looks good; please review my comment regarding the Kubernetes version update, as these two updates might not be compatible for now.

cbarria:
same as the other PR, we will set this to draft till we figure which version should be good to go.

jhoblitt (approved):

PR #818, opened 2023-05-08
(ccs) use lockmanager on all auxtel hosts

glennmorris:
Since it was already active on TTS and summit, this is adding BTS.


tony-johnson (approved):

jhoblitt (approved):
LGTM

PR #819, opened 2023-05-09
(node/auxtel-archiver.{tu,cp}.lsst.org) harmonize /data/allsky nfs export

jhoblitt:
BTS auxtel nfs exports are planned to be revised in #820


jhoblitt:
I'm converting this to a draft as we should add the allsky export to BTS and TTS as well.

cbarria (approved):
Looks Great 👍🏻

PR #821, opened 2023-05-09
(node/auxtel-archiver.{tu,cp}.lsst.org) rm /data/allsky nfs export

jhoblitt:
This nfs export was triggering a bug on auxtel-archiver.cp.lsst.org where when /data/allsky was exported as :/allsky, the :/data export would also become an export of /data/allsky.  As the :/data export already existed and had the same export client list as :/allsky, nfs client were already able to mount :/data/allsky.  It isn't known why this issue could not be reproduced on auxtel-archiver.tu.lsst.org nor as to why other exports of `/data/<sub dir>` do not trigger this issue.

The nfs-utils-1.3.0-mountd-root.patch patch included in nfs-utils-1.3.0-0.68.0.1.el7.2.src.rpm does not fix this issue.


cbarria (approved):
Looks Great! 👍🏻

PR #822, opened 2023-05-09
(role/ccs-mcm) merge role/atsccs into role/ccs-mcm 

jhoblitt:


glennmorris (approved):
Fine by me, given that auxtel-dc01.cp has been switched from atsccs to ccs-dc.

PR #823, opened 2023-05-11
(role/dnscache) convert from example42/network -> profile::nm

jhoblitt:
* Drops support for EL7 from dnscache role and removes the last usage of example42/network on EL8.
* Removes the CTIO backup network configuration from dns2.cp.lsst.org -- the virtual network interface is down and the configuration is known to be broken.
* This change has been manually tested on dns1.{cp,dev,ls,tu}.ls.lsst.org.


dtapiacl (commented):
Looks good to me, interesting stuff.

dtapiacl (approved):
Looks good to me.

PR #824, opened 2023-05-11
(spec) always skip testing on el7 instead of pinning tests to el8

jhoblitt:


dtapiacl (commented):
Looks good to me.

dtapiacl (approved):
Looks good.

PR #827, opened 2023-05-19
(core{01-03}.dev) Network Config and Tests

dtapiacl:


jhoblitt (commented):
It would be nice if the `(...)` prefix used on the commits was the same.  Another option would be to squash the two commits in this PR together. Other than that, this looks golden.

image

dtapiacl:
Updated the prefix for the commit to match the on of the PR. Let me know if this is correct.

cbarria (approved):
Looks Great 👍🏻

jhoblitt:
@dtapiacl The `(...)` section should indicate the hiera layer, so `(node/...)` would be better.  I have been meaning to put this policy into CI but haven't gotten to it yet.  Also, you are using both `add` and `Add` in commit messages. My opinion is that consistency is king.

jhoblitt (approved):
LGTM.

PR #828, opened 2023-05-22
(profile::ccs::common) add time package

jhoblitt:


glennmorris (approved):
Looks great, thanks.
At present we have duplicated packages: entries in comcam-ccs,yaml, lsstcam-ccs.yaml, etc - maybe they should be moved to this new place?

jhoblitt:
> At present we have duplicated packages: entries in comcam-ccs,yaml, lsstcam-ccs.yaml, etc - maybe they should be moved to this new place?

If the packages are intended to be installed on all ccs nodes, absolutely.

PR #831, opened 2023-05-23
(core/common) add packages and tests

dtapiacl:
Added screen package for installation. JQ Package is installed already under (profile::core::debugutils::packages)

### DIFF
--- a/site/profile/data/common.yaml
+++ b/site/profile/data/common.yaml
@@ -2,7 +2,9 @@
 profile::core::convenience::packages:
   - "ack"
   - "git"
+  - "jq"
   - "neovim"
+  - "screen"
   - "tree"
   - "vim"
 profile::core::debugutils::packages:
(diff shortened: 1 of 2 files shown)


site/profile/data/common.yaml
```diff
@@ -5,6 +5,7 @@ profile::core::convenience::packages:
   - "neovim"
   - "tree"
   - "vim"
+  - "screen"
```
jhoblitt: Please sort the packages list. The reason to keep lists sorted is to try to avoid merge conflicts down the road. It is impossible to completely avoid the problem but it is slightly less likely to happen with sorted lists. In vim/nvim this is easy to do with `<visual block>` then `!sort`.

jhoblitt (changes requested):
For the `(...)` block, I would suggest refer to the class that this change applies to, which would be `(profile::core::common)`.

Lets move `jq` from debugutils -> convenience so that it is present on all hosts.

Unit tests are needed.

dtapiacl:
1. Packages list was sorted with VIM. 
2. JQ was moved to convenience.
3. Unit tests were performed.

jhoblitt (approved):
LGTM

dtapiacl (commented):
Updates were made.

dtapiacl (commented):
Updates to files were made.

PR #833, opened 2023-05-30
(node/tel-hw1.ls.lsst.org) add /net/obs-env nfs mount

jhoblitt:


dtapiacl (approved):
Changes and tests look good to me. 👍🏻

PR #837, opened 2023-06-06
(gha) add arborist gha workflow

jhoblitt:
Note that this workflow is dependent on a secret added to this repo named "ARBORIST_GITHUB_ACTION". This secret is a   fine grained personal access token, which will expire on 2024-06-06 and need to be replaced.


cbarria (approved):
Looks Great 👍🏻

PR #840, opened 2023-06-15
 (node/auxtel-archiver.cp.lsst.org) rm all NFS exports; mount nfs-auxtel:/auxtel

jhoblitt:
The current `auxtel-archiver.cp.lsst.org` host has a bad CMOS battery and strange BIOS errors messages -- the host can no longer be trusted. This host dates from 2018 and support expired on `15 FEB 2023` and will be retired rather than repaired. The `/data` volume has been copied to `nfs-auxtel.cp.lsst.org:/auxtel` and all previous clients of `auxtel.archiver.cp` should already be updated to the new nfs export.  The host is being left in service temporarily as `ccs-ipa@auxtel-fp01.cp` is `scp`ing data to `/data`. In order to keep this process working, `nfs-auxtel.cp.lsst.org:/auxtel` is being mounted as `/data`.

### DIFF
--- a/hieradata/node/auxtel-archiver.cp.lsst.org.yaml
+++ b/hieradata/node/auxtel-archiver.cp.lsst.org.yaml
@@ -58,62 +58,8 @@ network::mroutes_hash:
     ensure: "absent"
     routes: {}
 
-# 139.229.146.0/24    antu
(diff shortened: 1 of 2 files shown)


cbarria (approved):
Looks Great 👍🏻

PR #842, opened 2023-06-19
(site/cp/cluster) bump auxtel and comcam ccs ts_sal version

glennmorris:
This is for XML 17.

After this change, this will be on TTS and summit, but not BTS. When this change also goes to BTS, this can be moved out of the site/ directory.


tony-johnson (approved):
Looks good. As we discussed we won't merge this to production until tomorrow so it does not accidentally get pushed to the summit too soon.

PR #843, opened 2023-06-20
(dev/role) set rke version 1.4.6-rc4

cbarria:
we need to bump the version of the rke client in order to crank that cluster.yaml from lsst-it/k8s-cookbook#178 to version ```v1.25.9-rancher2-2``` that matches rancher v2.7.4

### DIFF
--- a/spec/hosts/roles/rke_spec.rb
+++ b/spec/hosts/roles/rke_spec.rb
@@ -25,6 +25,7 @@ describe "#{role} role" do
   # rubocop:disable Naming/VariableNumber
   on_supported_os.merge('almalinux-9-x86_64': alma9).each do |os, facts|
     # rubocop:enable Naming/VariableNumber
+
     context "on #{os}" do
       let(:facts) { facts }
       let(:node_params) do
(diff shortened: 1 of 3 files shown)


jhoblitt (changes requested):
I feel like this should be covered by spec. Could you add an example to `spec/hosts/roles/rke_spec.rb` for the `dev` site?

spec/hosts/roles/rke_spec.rb
```diff
@@ -34,19 +35,69 @@
         }
       end
 
-      lsst_sites.each do |site|
-        fqdn = "#{role}.#{site}.lsst.org"
-        override_facts(facts, fqdn: fqdn, networking: { fqdn => fqdn })
+      describe 'rke.tu.lsst.org', :sitepp do
```
cbarria: this is weird...

spec/hosts/roles/rke_spec.rb
```diff
@@ -34,19 +35,69 @@
...
+        let(:site) { 'tu' }
+
```
cbarria, line 44: this probably needs a new fact request? because of 'override_facts'?

cbarria:
i know there is something weird with the test... but the idea is there... tried to remove the ```site.each do``` and replace it with an explicit definition of the site with ```let(:site) { 'dev' }```

spec/hosts/roles/rke_spec.rb
cbarria: when change host test, needs a new refact?

spec/hosts/roles/rke_spec.rb
jhoblitt: The `        override_facts(facts, fqdn: fqdn, networking: { fqdn => fqdn })` is still going to be needed for each site.

cbarria: i used ```facts.merge!(fqdn: 'rke.tu.lsst.org')``` like the forman role tests, or its better to use the override_facts?

jhoblitt: `networking: { fqdn => fqdn })` is still needed.

jhoblitt (approved):
LGTM.

PR #844, opened 2023-06-20
(core{01-03}.dev) Network Config / Variant / Tests

dtapiacl:


jhoblitt (changes requested):
Cut'n'pasting the same config for multiple hosts is a sign that common config needs to be factored out. The existing configuration being used for EL9 dev hypervisors is at `site/dev/role/hypervisor/osfamily/RedHat/major/9.yaml`.  These changes should probably either be included there (which would be problematic as this role is being used for multiple hardware configurations) or replaced with a `site/dev/variant/*` .

jhoblitt (changes requested):
The variant looks reasonable but tests are needed (note that the PR and commit title says config and tests :smile: ).

As these are critical hosts, we should add examples to all 3 of the nodes.  Here is an existing `hosts/nodes` that tests variant networking that should be a reasonable template:

https://github.com/lsst-it/lsst-control/blob/production/spec/hosts/nodes/lsstcam-dc01.ls.lsst.org_spec.rb

hieradata/site/dev/variant/1114S-rev_1.01.yaml
```diff
@@ -0,0 +1,44 @@
+---
+profile::nm::connections:
+  enp1s0f1:
```
jhoblitt, line 3: Please add spec on each host for these interfaces.

jhoblitt (changes requested):

jhoblitt:
Note that https://github.com/lsst-it/lsst-control/pull/868 has been merged and the hiera key `profile::nm::connections` is no longer being  used / will need to be updated.

hieradata/site/dev/variant/1114S-rev_1.01.yaml
dtapiacl: Hi @jhoblitt, I added the spec for each of the interfaces but I'm having issues with the nm connections. I would appreciate your help if you could help revise the code based on the logs below.

dtapiacl: Updated and fixed tests. PR should be mergeable now.

jhoblitt (approved):
LGTM. Thanks for persisting.

PR #845, opened 2023-06-20
(cluster/auxtel-ccs, comcam-ccs) bump ts_sal version for cycle 31

glennmorris:
This moves the setting from being tu and cp site-specific to global, ie also on the bts.


glennmorris:
@tony-johnson for BTS cycle 31: https://lsstc.slack.com/archives/C02NC12G946/p1687291201194819

tony-johnson (approved):
Looks good to me.

PR #846, opened 2023-06-20
(node/core{1-3}.ls) add networking

dtapiacl:
Added FQDN network config for core ls nodes.


jhoblitt (changes requested):
Please add spec examples.

PR #848, opened 2023-06-21
(cluster/pillan) add ip rules for br3035

jhoblitt:
As with on the manke cluster, I suspect the ip rules are not matching anything and are being bypassed by nftables, which is why disabling `rp_filter` is necessary.


cbarria (approved):
Looks Great 👍🏻

PR #859, opened 2023-06-26
(role/rke) fs.inotify.max_inotify_instances=104857

jhoblitt:
To resolve this error message from kubectl when tailing pod logs:

    failed to create fsnotify watcher: too many open files


cbarria (approved):
Looks Great 👍🏻

PR #861, opened 2023-06-28
(node/core*) statically assign BMC IP address

jhoblitt:
This resolves the bug introduced in #728 / #729 that forced all BMCs to be configured with DHCP, including the core* nodes. The core nodes should have the BMC IP statically assigned as these are the hypervisors that host the per site DHCP server.


cbarria (approved):
Looks Great, thanks a lot! 👍🏻

PR #864, opened 2023-07-03
(role/hypervisor) fix libvirt configuration on EL >= 9.2

jhoblitt:
It appears that under EL9.2, the systemd units/sockets for libvirt have changed.

Network configuration was also removed from the dev hypervisor role to allow testing with a VM. EL7 only networking configuration remains for the cp, ls, and tu sites as refactoring runs the risk of an unplanned networking restart.  The EL7 networking configuration should be removed as the core nodes are migrated to EL9.


Puppetfile
dtapiacl, line 1: It was some versions behind.

jhoblitt: It wasn't very far behind and the version change isn't strictly needed. However, I usually update module versions at the same time I'm testing other changes.

dtapiacl (approved):
Thank you @jhoblitt. Changes look good to me and are easy to understand.

PR #865, opened 2023-07-04
(site/dev) assign static IP addresses to all dns instances; switch hosts to use all dev dns servers

jhoblitt:


dtapiacl (approved):
Looks good to me 👍🏻

PR #866, opened 2023-07-06
(rancher.dev) dhcp on ens3 interface and tests

cbarria:
basic network config and test for the rancher.dev cluster

### DIFF
--- a/hieradata/cluster/rancher.yaml
+++ b/hieradata/cluster/rancher.yaml
@@ -2,8 +2,27 @@
 # Be advised that there is a rancher cluster per site and changes here will
 # apply to all of them.
 clustershell::groupmembers:
-  kueyen: {group: "rancher", member: "rancher[1-3]"}
+  rancher: {group: "rancher", member: "rancher[1-3]"}
 
 profile::core::common::manage_powertop: true
 profile::core::ospl::enable_rundir: true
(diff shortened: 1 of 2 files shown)


hieradata/cluster/rancher/osfamily/RedHat/major/8.yaml
```diff
@@ -0,0 +1,19 @@
+---
+profile::nm::connections:
```
jhoblitt: I'm not sure if this should be limited to EL8 as the network config should be the same for EL9.

cbarria: where or how in the hierarchy can i put this to grab Alma 8 and 9, but not Centos 7? (i could copy and create a 9.yaml, but the idea is not to repeat the code right?)

jhoblitt: There is no place in hierarchy which would apply this to EL7 as `profile::nm` is only included if the `os.release.major` is NOT `7`.

cbarria: great! so i can go higher in the hierarchy and not affect EL7, thanks!

jhoblitt (approved):
LGTM.

PR #868, opened 2023-07-07
(*) convert profile::nm -> nm

jhoblitt:
Note that the `no-auto-default=*` `NetworkManager.conf` directive is intentionally being removed as part of the conversion to the external `nm` module.


cbarria (approved):
Looks Great, Thanks! 👍🏻

dtapiacl (approved):
Looks good 👍🏻

PR #874, opened 2023-07-27
(profile::ccs::common) move packages here from ccs cluster files

glennmorris:
This avoids duplicating the same information.


jhoblitt (approved):
LGTM.

PR #876, opened 2023-07-28
(ayekan.ls) Initial network config and role

cbarria:
This is basic initial configuration and test for the Observability cluster.
no tagging on VLANs for the secondary interface for now, will add if needed.


dtapiacl (approved):
Looks good. Nice catch on the nm::connections.

PR #879, opened 2023-07-31
(cluster/luan) rm EL7 network config

jhoblitt:
No long user as luan has been migrated to EL9. No vlan interfaces are in use and no manual network configuration is currently required.


dtapiacl (approved):
LGTM.

PR #880, opened 2023-08-01
(ccs) install file-transfer utilities

glennmorris:
Adding this to site/profile/ccs since it already has the machinery for installing files from nexus

NB this requires https://github.com/lsst-it/lsst-puppet-hiera-private/pull/90

### DIFF
--- a/hieradata/role/atsdaq.yaml
+++ b/hieradata/role/atsdaq.yaml
@@ -7,3 +7,5 @@ classes:
   - "profile::core::debugutils"
   - "profile::core::nfsclient"
   - "profile::core::nfsserver"
+
+profile::ccs::file_transfer::install: true
--- a/hieradata/role/comcam-fp.yaml
+++ b/hieradata/role/comcam-fp.yaml
@@ -8,3 +8,5 @@ classes:
   - "profile::core::debugutils"
   - "profile::core::nfsclient"
   - "profile::core::nfsserver"
+
+profile::ccs::file_transfer::install: true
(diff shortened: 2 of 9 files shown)


jhoblitt (commented):
@glennmorris This looks reasonable.  Could we have at least some minimal spec coverage for the new class?

glennmorris:
> @glennmorris This looks reasonable. Could we have at least some minimal spec coverage for the new class?

Something beyond the following?

https://github.com/lsst-it/lsst-control/pull/880/commits/668d8a3efc3cae89a64e7a15cd05b1c5b5315112
https://github.com/lsst-it/lsst-control/pull/880/commits/d2785d2b3e738f5adb70f58fa6ddc2220521fd22

jhoblitt:
> > @glennmorris This looks reasonable. Could we have at least some minimal spec coverage for the new class?
> 
> Something beyond the following?
> 
> [668d8a3](https://github.com/lsst-it/lsst-control/commit/668d8a3efc3cae89a64e7a15cd05b1c5b5315112) [d2785d2](https://github.com/lsst-it/lsst-control/commit/d2785d2b3e738f5adb70f58fa6ddc2220521fd22)

There aren't any class level tests at all.  I do prefer the more functional tests of the results rather than all of the logic but it looks like there's a number of script files being installed that don't have coverage.

glennmorris:
I've added some [tests](https://github.com/lsst-it/lsst-control/blob/IT-4662/ccs-file-transfer/spec/classes/ccs/file_transfer_spec.rb) for the class.

jhoblitt (approved):
This looks much improved. Thanks for putting in the effort. :+1:

PR #882, opened 2023-08-11
(common) rm /etc/rc.d/rc.local on EL*

jhoblitt:
This file is unused and EL9 complains when it exists but is not
executable:

    [155682.912720] systemd-rc-local-generator[56182]: /etc/rc.d/rc.local is not marked executable, skipping.
    [155683.394680] systemd-rc-local-generator[56208]: /etc/rc.d/rc.local is not marked executable, skipping.
    [155683.823637] systemd-rc-local-generator[56231]: /etc/rc.d/rc.local is not marked executable, skipping.


glennmorris (approved):
Fine by me. (You think removing the file is better than making it executable?)

jhoblitt:
> Fine by me. (You think removing the file is better than making it executable?)

That's another option.  I'm going to leave this unmerged for a little bit as I think as, ironically, I think we might need to use `rc.local` as a kludge for the summit lsstcam nodes.  I'll explain via slack.

jhoblitt:
Rebased on current master.

PR #883, opened 2023-08-21
(node/netdev.dev) rm; host does not exist

jhoblitt:


cbarria (approved):
Looks Great 👍🏻 Thanks for cleaning things up

PR #892, opened 2023-08-24
(role/nfsserver.cp) add nfs2 to jhome client list

cbarria:
In order to make nfs1 backups more easy, we added access network to the jhome mount on the 139.229.165/24 network.

### DIFF
--- a/hieradata/site/cp/role/nfsserver.yaml
+++ b/hieradata/site/cp/role/nfsserver.yaml
@@ -36,6 +36,7 @@ nfs::nfs_exports_global:
       %{facts.networking.ip}/32(ro,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.146.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.160.0/24(rw,nohide,insecure,no_subtree_check,async,no_root_squash)
+      nfs2.cp.lsst.org(rw,nohide,insecure,no_subtree_check,async,no_root_squash)
   /data/lsstdata:
     clients: >-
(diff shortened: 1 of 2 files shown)


hieradata/site/cp/role/nfsserver.yaml
```diff
@@ -36,6 +36,7 @@ nfs::nfs_exports_global:
       %{facts.networking.ip}/32(ro,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.146.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.160.0/24(rw,nohide,insecure,no_subtree_check,async,no_root_squash)
+      139.229.165.0/24(rw,nohide,insecure,no_subtree_check,async,no_root_squash)
```
jhoblitt: Does the entire subnet need to be allowed?  I thought it only needed to be mounted on a single host?

cbarria: i asked Cristian and said the entire 160 segment is OK, but i see your point... will check back on this 👌🏻

jhoblitt: It is better to allow specific hosts, preferably by fqdn.

csilva-cl: This is following the config of the other mounts, but seems reasonable to create approvals for  specific hosts to prevent unnecessary spread of nfs. That would require significant work, so I would approve this PR to get the config in then we can re-plan the config. of the nfs

csilva-cl (approved):
approved as per my comment

PR #897, opened 2023-08-28
(foreman.cp) Fix range on LHN summit network

cbarria:
The foreman DHCP was having IP conflicts with the MetalLB Pool on Yagan... so in order to avoid this issue, we moved the range on the foreman DHCP over the range of IPs reserved for MetalLB. (yagan and chonchon use LHN)


jhoblitt (approved):
LGTM.

PR #898, opened 2023-08-29
(ccs) update ts_sal_utils for cycle 32

glennmorris:
This makes ts_sal_utils 7.4.0-1 the default on all active ccs clusters. Remove now superfluous site-specific settings.


tony-johnson (approved):
Looks good to me.

PR #905, opened 2023-09-01
(ccs) allow installation of required jdk8 on el9

glennmorris:
The jdk8 rpm required by CCS does not install out of the box on EL9 due to the merging of /bin and /usr/bin making it think some dependencies are not present (when they are). Install a trivial compat rpm to provide the "missing" dependencies and allow jdk8 to install.

### DIFF
--- a/site/profile/manifests/ccs/common.pp
+++ b/site/profile/manifests/ccs/common.pp
@@ -31,6 +31,10 @@ class profile::ccs::common (
     include profile::daq::sysctl
   }
 
+  if fact('os.release.major') == '9' {
+    include profile::ccs::el9
+  }
+
   include ccs_software
   include java_artisanal
   include java_artisanal::java17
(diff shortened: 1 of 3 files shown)


tony-johnson (commented):
The goal looks good to me, but I don't know enough about RPMs to be able to tell if it will work as expected or not. We could configure one of the summit lsstcam-dcNN nodes to use this, if that would be useful for verifying it.

glennmorris:
> The goal looks good to me, but I don't know enough about RPMs to be able to tell if it will work as expected or not. We could configure one of the summit lsstcam-dcNN nodes to use this, if that would be useful for verifying it.

I already used it on all lsstcam-dc EL9 nodes, and it worked fine.

tony-johnson (approved):
Approved, based on the fact that it appeared to work fine on lsstcam-dc01.cp.lsst.org.

PR #909, opened 2023-09-07
(chango) add network config and role

dtapiacl:


jhoblitt (approved):

hieradata/cluster/chango/role/rke.yaml
```diff
@@ -0,0 +1,6 @@
+---
+classes:
+  - "profile::core::sysctl::rp_filter"
+profile::core::sysctl::rp_filter::enable: false
+profile::core::rke::enable_dhcp: true
```
jhoblitt: Everything above this line isn't needed on chango.

hieradata/cluster/chango/role/rke.yaml
```diff
@@ -0,0 +1,6 @@
...
+profile::core::rke::version: "1.4.6-rc4"
```
jhoblitt: This single line can be moved to `cluster/chango.yaml` and then this file could be removed completely.

hieradata/cluster/chango.yaml
```diff
@@ -0,0 +1,21 @@
+---
+clustershell::groupmembers:
+  chango: {group: "chango", member: "chango[01-03]"}
+cni::plugins::checksum: "f3a841324845ca6bf0d4091b4fc7f97e18a623172158b72fc3fdcdb9d42d2d37"
+cni::plugins::enable: ["macvlan"]
+cni::plugins::version: "1.2.0"
+profile::core::common::manage_powertop: true
+profile::core::ospl::enable_rundir: true
```
jhoblitt: This isn't needed, it is only for running DDS.

hieradata/cluster/chango.yaml
```diff
@@ -0,0 +1,21 @@
+---
+clustershell::groupmembers:
+  chango: {group: "chango", member: "chango[01-03]"}
+cni::plugins::checksum: "f3a841324845ca6bf0d4091b4fc7f97e18a623172158b72fc3fdcdb9d42d2d37"
+cni::plugins::enable: ["macvlan"]
+cni::plugins::version: "1.2.0"
```
jhoblitt: As there are no users of multus on this cluster, all of the `cni::*` keys should be removed.

jhoblitt (changes requested):

hieradata/cluster/chango/role/rke.yaml
dtapiacl: Cleaned up and updated.

hieradata/cluster/chango/role/rke.yaml
dtapiacl: `profile::core::rke::version: "1.4.6-rc4"` was moved to cluster/chango.yaml

hieradata/cluster/chango.yaml
dtapiacl: Cleaned up.

hieradata/cluster/chango.yaml
dtapiacl: Removed from configuraiton.

hieradata/cluster/chango.yaml
```diff
@@ -0,0 +1,17 @@
+---
+clustershell::groupmembers:
+  chango: {group: "chango", member: "chango[01-03]"}
+profile::core::rke::version: "1.4.6-rc4"
+nm::connections:
```
jhoblitt: In the scenario where there is only a single interface and its using dhcp, no network configuration is needed as NM defaults to being an dhcp client on active links.  The best configuration is usually no-configuration.

jhoblitt (changes requested):

hieradata/cluster/chango.yaml
dtapiacl: Network config was removed as advised.

jhoblitt (approved):

PR #910, opened 2023-09-07
(ccs) make java17 the default for tu/comcam

glennmorris:


tony-johnson (approved):
Looks good to me. Already tested on TTS,

PR #929, opened 2023-09-27
(node/core0[123].cp.lsst.org) add 1114s based servers

jhoblitt:
As replacement for the core[123].cp.lsst.org r440s.


cbarria (approved):
Looks Great 👍🏻

PR #930, opened 2023-09-28
(node/hexrot) profile for hexrot

cbarria:
This is a request on IT-4666 to:

- install anaconda and setup virtual environment with Python 3.11 (default for all users)
- install python packages on that environment to run ts_m2gui package
- download ts_config_mttcs repo on /opt/ts_config_mttcs
- change ownership and permissions to the /rubin/ folder (recursive) to the saluser
- create a softlink to /rubin/mtm2/python/run_m2gui
- set a couple of environment Variables to work with

### DIFF
--- a/site/profile/manifests/ts/hexrot.pp
+++ b/site/profile/manifests/ts/hexrot.pp
@@ -1,0 +1,45 @@
+# @summary
+# Sets up repos and symlinks for hexrot
+
+class profile::ts::hexrot {
+  vcsrepo { '/opt/ts_config_mttcs':
+    ensure             => present,
+    provider           => git,
+    source             => 'https://github.com/lsst-ts/ts_config_mttcs.git',
+    revision           => 'v0.12.8',
+    keep_local_changes => false,
+  }
+  file { '/etc/profile.d/hexrot_path.sh':
+    ensure  => file,
+    mode    => '0644',
+    # lint:ignore:strict_indent
+    content => @(ENV),
+        export QT_API="PySide6"
+        export PYTEST_QT_API="PySide6"
+        export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+        | ENV
+    # lint:endignore
+    require => Vcsrepo['/opt/ts_config_mttcs'],
+  }
+  file { '/rubin/mtm2/python':
+    ensure => directory,
+    owner  => 73006,
+    group  => 73006,
+  }
+  file { '/rubin/mtm2/python/run_m2gui':
+    ensure => link,
+    owner  => 73006,
+    group  => 73006,
+    target => '/opt/anaconda/envs/py311/bin/run_m2gui',
+  }
+  file { ['/rubin/rotator', '/rubin/hexapod', '/rubin/mtm2']:
+    ensure  => directory,
+    owner   => 73006,
+    group   => 73006,
+    recurse => true,
+  }
+  file { ['/rubin/rotator/log', '/rubin/hexapod/log', '/rubin/mtm2/log']:
+    ensure => directory,
+    mode   => '0775',
+  }
+}
--- a/spec/hosts/roles/hexrot_spec.rb
+++ b/spec/hosts/roles/hexrot_spec.rb
@@ -6,6 +6,8 @@ role = 'hexrot'
 
 describe "#{role} role" do
   on_supported_os.each do |os, os_facts|
+    next unless os =~ %r{almalinux-9-x86_64}
+
     context "on #{os}" do
       let(:facts) { os_facts }
       let(:node_params) do
@@ -24,24 +26,117 @@ describe "#{role} role" do
 
           it { is_expected.to compile.with_all_deps }
 
+          include_examples 'debugutils'
           include_examples 'common', os_facts: os_facts, site: site
           include_examples 'x2go packages', os_facts: os_facts
+          include_examples 'ni_packages'
+          include_examples 'nexusctio'
           it { is_expected.to contain_class('mate') }
 
           # XXX hexrot uses devicemapper, so the docker example group isn't included
-          it { is_expected.to contain_class('docker') }
+          it do
+            is_expected.to contain_class('docker').with(
+              package_source: 'docker-ce',
+              socket_group: 70_014,
+              socket_override: false,
+              storage_driver: 'devicemapper',
+            )
+          end
 
-          %w[
-            profile::core::common
-            profile::core::debugutils
-            profile::core::docker
-            profile::core::docker::prune
-            profile::core::ni_packages
-          ].each do |c|
-            it { is_expected.to contain_class(c) }
+          it { is_expected.to contain_cron__job('docker_prune') }
+
+          it do
+            is_expected.to contain_vcsrepo('/opt/ts_config_mttcs').with(
+              ensure: 'present',
+              provider: 'git',
+              source: 'https://github.com/lsst-ts/ts_config_mttcs.git',
+              revision: 'v0.12.8',
+              keep_local_changes: 'false',
+            )
+          end
+
+          pkgs = {
+            'pyside6' => {
+              'channel' => 'conda-forge',
+              'version' => '6.7.0',
+            },
+            'qasync' => {
+              'channel' => 'conda-forge',
+              'version' => '0.23.0',
+            },
+            'qt6-charts' => {
+              'channel' => 'conda-forge',
+              'version' => '6.7.0',
+            },
+            'ts-m2com' => {
+              'channel' => 'lsstts',
+              'version' => '1.5.4',
+            },
+            'ts-m2gui' => {
+              'channel' => 'lsstts',
+              'version' => '1.0.2',
+            },
+          }
+
+          it do
+            is_expected.to contain_class('anaconda').with(
+              anaconda_version: 'Anaconda3-2023.07-2',
+              python_env_name: 'py311',
+              python_env_version: '3.11',
+              conda_packages: pkgs,
+            )
           end
 
           it { is_expected.to contain_package('docker-compose-plugin') }
+
+          it do
+            is_expected.to contain_file('/etc/profile.d/hexrot_path.sh').with(
+              ensure: 'file',
+              mode: '0644',
+              content: <<~CONTENT,
+              export QT_API="PySide6"
+              export PYTEST_QT_API="PySide6"
+              export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+              CONTENT
+            )
+          end
+
+          it do
+            is_expected.to contain_file('/rubin/mtm2/python').with(
+              ensure: 'directory',
+              owner: '73006',
+              group: '73006',
+            )
+          end
+
+          it do
+            is_expected.to contain_file('/rubin/mtm2/python/run_m2gui').with(
+              ensure: 'link',
+              owner: '73006',
+              group: '73006',
+              target: '/opt/anaconda/envs/py311/bin/run_m2gui',
+            )
+          end
+
+          ['/rubin/rotator', '/rubin/hexapod', '/rubin/mtm2'].each do |path|
+            it do
+              is_expected.to contain_file(path).with(
+                ensure: 'directory',
+                owner: '73006',
+                group: '73006',
+                recurse: 'true',
+              )
+            end
+          end
+
+          ['/rubin/rotator/log', '/rubin/hexapod/log', '/rubin/mtm2/log'].each do |path|
+            it do
+              is_expected.to contain_file(path).with(
+                ensure: 'directory',
+                mode: '0775',
+              )
+            end
+          end
         end # host
       end # lsst_sites
     end # on os
--- a/Puppetfile
+++ b/Puppetfile
@@ -12,6 +12,7 @@ mod 'example42/network', git: 'https://github.com/lsst-it/puppet-network', ref: 
 mod 'jamtur01/httpauth', '0.0.6'
 mod 'jcpunk/irqbalance', '1.0.5'
 mod 'jhoblitt/ipmi', '5.3.1'
+mod 'lsst/anaconda', '0.1.0'
 mod 'lsst/ccs_daq', '1.1.0'
 mod 'lsst/ccs_database', '3.3.0'
 mod 'lsst/ccs_hcu', '3.5.0'
--- a/hieradata/role/hexrot.yaml
+++ b/hieradata/role/hexrot.yaml
@@ -1,5 +1,6 @@
 ---
 classes:
+  - "anaconda"
   - "mate"
   - "profile::core::common"
   - "profile::core::debugutils"
@@ -7,6 +8,7 @@ classes:
   - "profile::core::docker::prune"
   - "profile::core::ni_packages"
   - "profile::core::x2go"
+  - "profile::ts::hexrot"
   - "profile::ts::nexusctio"
 files:
   /rubin:
@@ -18,6 +20,8 @@ files:
     subscribe:
       - "Package[runHexEui]"
       - "Package[runRotEui]"
+      - "Package[runM2Cntlr]"
+
 # / on hexrot.cp was formated with xfs fstype=0 (long, long, long ago) and is
 # not compatible with overlayfs[2]
 profile::core::docker::storage_driver: "devicemapper"
@@ -27,3 +31,23 @@ accounts::group_list:
     ensure: "present"
     gid: 70014
     forcelocal: true
+
+anaconda::anaconda_version: "Anaconda3-2023.07-2"
+anaconda::python_env_name: "py311"
+anaconda::python_env_version: "3.11"
+anaconda::conda_packages:
+  pyside6:
+    channel: "conda-forge"
+    version: "6.7.0"
+  qasync:
+    channel: "conda-forge"
+    version: "0.23.0"
+  qt6-charts:
+    channel: "conda-forge"
+    version: "6.7.0"
+  ts-m2com:
+    channel: "lsstts"
+    version: "1.5.4"
+  ts-m2gui:
+    channel: "lsstts"
+    version: "1.0.2"
--- a/spec/support/spec/nexusctio.rb
+++ b/spec/support/spec/nexusctio.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+shared_examples 'nexusctio' do
+  it do
+    is_expected.to contain_yumrepo('nexus-ctio').with(
+      'baseurl' => 'http://cagvm3.ctio.noao.edu/nexus/repository/labview-rpm/rubin/',
+      'gpgcheck' => false,
+    )
+  end
+
+  it do
+    is_expected.to contain_file('/etc/pki/rpm-gpg/RPM-GPG-KEY-MSO').with(
+      ensure: 'file',
+      mode: '0644',
+      content: <<~GPG,
+          -----BEGIN PGP PUBLIC KEY BLOCK-----
+        Version: GnuPG v2.0.22 (GNU/Linux)
+
+        mQINBGC3zR8BEACw7AKpnTXmhvjHU5hfdBfbbdOx8azUtW3Oogr6zUcAMd0meTfj
+        4msB71LePoMX6WskN9RckoIJvxlIjcVPlbCe7euBDLiQL1KhuQXWK5veYIlsBYMx
+        agL8CA/rHEk1yfvuk9pxS7CqziRQ4laxJPnYbrOaRZevYlumyusDfGQ17x3EbM1D
+        N3UYvnzhla0lq/YmGSLVNgmqMV8PJSzv+iQku5MbI4ahxSCnVedOvepXdcePQPwf
+        /IZc/4IwquKjTNXCs+WhlMbM7T/0M1WeYm7h52UwUOlWB2pgN1ryW+iSbLCu0rBd
+        XJtPw4iERbhKpZmkmZJ6vrpCA+E2OFW2LIFve0N5mAUY0JkliFE34kG+FY6ZtD7z
+        vlF4U0AxzAOvo4zoIJcxRHjt7u4UsOIpMbBvm+B1rtbX3ZGGiZecpFhe5GONrh++
+        jvT4kcpKjADY4rSq583iDZ/mWF+XUozyTmqbuHxFVTjAKCNvQE0WyL9+5g9zCH1a
+        7AtIOSrfVP5YBHfLMiPOXpep8Pf7CaNAbSQb5JxFABYSRsLM6v2+Y3zm61dSIfeU
+        ZTRIFAsWG0Db5VdtdznBZnZRa5Avs6AUN9KOiVWi/D1KofZeeHU6DSfW+4Wq8vr4
+        Ekm8YUfQuPP6Fvre5qvytBf17MGg0vKZ7Z/ipdhs2StxFnoO+3yK2BfoKwARAQAB
+        tF9HUEcgTVNPIChHUEcgTVNPIHNpZ25hdHVyZSBmb3Igc3RhYmxlIHJwbSBmaWxl
+        cyBpbiBOZXh1czMgcmVwb3NpdG9yeSkgPGRpZWdvLmdvbWV6QG5vaXJsYWIuZWR1
+        PokCOQQTAQIAIwUCYLfNHwIbAwcLCQgHAwIBBhUIAgkKCwQWAgMBAh4BAheAAAoJ
+        EM5tAOEzBhcslQgP/2uGa/PZwVSo9/PHvnr7sD6/QT4hM0/Z2ngCGo0xOTTLm89C
+        K0DxkU21L9m2/BUZPJp4krB5p5GOUjiRdpouiPElnULBaBB8eZIVQWK3ifenqVq6
+        FqxNqFDBKcKQ9ZVgvz88GYHLDSYTvv2QtTozCVhjcSWCPorYRuLbliOc4k7jqmjk
+        5xUlCLcaGyHYZNXcLwDztgioF4AXyL3NUq2wySLfukJ0VxZ4z5jBfx5u47jUzvOV
+        j/t/OHZTwL4UGziEXgTb0j8ZRt2QciA5Sh0oXGUIlqZhWY88TBeG9xKUcsxB6TpV
+        NUSLOFt7Vq1Gxi9Eg8TWHvHyRRG/pfWIQyJdX5dKcAzsZMmB3jsx5TypnbxwM5dQ
+        k+IRCCLQQeLgBupBRKlF4PLns7R/CzouFIo9J/pVcxsZS2r2+t/S2i1N6/ApuqGR
+        Zo1QcKP8LGklwzcM6WQUHNlFJMzudULxAPt3ZNgKtktxEzfF44u9jWediWCHZv8N
+        /QgfxfaHMt4cQG2FCfZbISeBibo2YpRzPxMmuuTrS9tUEQ9VLiy59keZtu9hzD8R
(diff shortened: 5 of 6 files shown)


site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,80 @@
...
+# @param python_env_name
+#   if exist, sets the name of the virtual environment
+#
+# @param python_env_version
+#   if exist, sets the version of python for the virtual environment
+#
+# @param conda_packages
+#   if contains at least 1 tuple (name and channel), will install the package
+#    
+class profile::core::anaconda (
```
jhoblitt: This is functionality that would be better as an external module instead of a profile.

Did you try the puppet/python module?  It claims to have anaconda support: https://github.com/voxpupuli/puppet-python/blob/9dcd35b4bda7247c03f0ad8f1142db3d30f207ce/manifests/install.pp#L148-L164

It doesn't look like any of the acceptance tests cover anaconda, so it may or may not need some repair.  I would probably start by writing acceptance tests for anaconda.

One of the large advantages of using puppet/python is that the `python::pip` defined type can be used. https://github.com/voxpupuli/puppet-python/blob/9dcd35b4bda7247c03f0ad8f1142db3d30f207ce/manifests/pip.pp#L51

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,38 @@
...
+    owner   => 73006,
+    group   => 73006,
+    require => Class[profile::core::anaconda],
+  }
+  file { '/rubin/mtm2/python/run_m2gui':
+    ensure  => link,
+    owner   => 73006,
+    group   => 73006,
+    target  => '/opt/anaconda/envs/py311/bin/run_m2gui',
+    require => File['/rubin/mtm2/python'],
```
jhoblitt: This require isn't necessary as the `file` type will "autorequire" parent directories. https://github.com/puppetlabs/puppet/blob/main/lib/puppet/type/file.rb#L387-L401

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,38 @@
...
+        | ENV
+    # lint:endignore
+  }
+  file { '/rubin/mtm2/python':
+    ensure  => 'directory',
+    owner   => 73006,
+    group   => 73006,
+    require => Class[profile::core::anaconda],
+  }
+  file { '/rubin/mtm2/python/run_m2gui':
```
jhoblitt, line 29: Does this file need to be executable?  If `mode` isn't set on a `file` resource it defaults to https://github.com/puppetlabs/puppet/blob/3c2306b494e5378d52447f192c5f84f754feeb76/lib/puppet/util.rb#L582

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,80 @@
...
+  String $python_env_name,
+  String $python_env_version,
+  Tuple   $conda_packages,
+) {
+  case $facts['os']['family'] {
+    'RedHat': {
+      #package { 'wget':
```
jhoblitt: I realize this is commented out but as an FYI, it is usually preferable to use the `ensure_packages()` function to install packages that might also have a package resource declared by a module/class/define.

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,38 @@
...
+    content => @(ENV),
+        #!/usr/bin/bash
+        export QT_API="PySide2"
+        export PYTEST_QT_API="PySide2"
+        export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+        | ENV
+    # lint:endignore
+  }
+  file { '/rubin/mtm2/python':
+    ensure  => 'directory',
```
jhoblitt: It looks like we have been inconsistent about quoting or not-quoting file resource's ensure parameter value.  We should probably pick a style and keep with it.  It working without being quoted (as a bare-word) is a left over from the early days of puppet.  I've had some discussion with other puppet people and there is disagreement.  The official puppet style guide says not to quote enum values: https://www.puppet.com/docs/puppet/8/style_guide.html#style_guide_module_design-quoting ... so we could go with that for now. What do you think?

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,15 +32,26 @@
...
+            mate
+            profile::core::anaconda
             profile::core::common
             profile::core::debugutils
             profile::core::docker
             profile::core::docker::prune
             profile::core::ni_packages
+            profile::core::x2go
+            profile::ts::hexrot
+            profile::ts::nexusctio
```
jhoblitt: So testing directly for the inclusion of other profile classes is actually a bit of an anti-pattern.  The reason this is in here is there were no tests at all and I banged something out in a couple of minutes.  Testing for external module is fine.  Testing directly for profile classes makes it hard to refactor the profiles.  This is acceptable for the moment but its technical debt.

site/profile/manifests/ts/hexrot.pp
cbarria: thats great, i was just trying to avoid "corrective" puppet execution, i will remove the require, thanks!

site/profile/manifests/ts/hexrot.pp
cbarria: this is weird, the original file has permissions 755 and the symlink gets created with permissions 777, so it works in this case, but doesn't default to 644 🤔 (and the file doesn't exist previously)
`lrwxrwxrwx 1 saluser 73006 38 Sep 27 19:21 run_m2gui -> /opt/anaconda/envs/py311/bin/run_m2gui
(py311) -bash-4.2$ ls -la /opt/anaconda/envs/py311/bin/ | grep run_m2gui
-rwxr-xr-x  1 root root      272 Sep 27 19:29 run_m2gui`

site/profile/manifests/core/anaconda.pp
cbarria: yes, that's exactly why is commented out, the `ni_packages` already has this packages, so it conflicts out.
i will change it to `ensure_packages()`

site/profile/manifests/ts/hexrot.pp
cbarria: i thought that single quotes meant literal string (no resolving variables or computing inside) and double quotes when you need to resolve or insert code inside, for example this: (was literally breaking my head)
`content => @("SOURCE"),
          #!/usr/bin/bash
          source /opt/anaconda/bin/activate ${python_env_name}
          | SOURCE`
 needed to resolve that variable there and was always coming out literal until i read this :
 [https://www.puppet.com/docs/puppet/7/lang_data_string.html#lang_data_string_heredocs-interpolation](url)  
 and notice that to be able to interpolate inside a Heredoc, dobles quotes need to be put around the NAME...

site/profile/manifests/core/anaconda.pp
cbarria: yes, my first attempt of doing this was using the python module, but always had issues installing and later on, resolving or trying to install packages with Conda... (didn't resolve any dependencies and took an hour to try to install Te-Wei packages with no success, puppet always timed out)
couldn't make it work like i needed to set every dependency... (they have a 17GB docker image with way more dependencies that this needs, so picking them was a nightmare.) was almost throwing the towel till i find this :
[https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community](url)
so i tried going back to Basic and installing everything like anaconda manual says and it worked great and fast once the new solver was set 👍🏻 (no dependency issues or manually having to pick them)

site/profile/manifests/ts/hexrot.pp
jhoblitt: Oh, this makes sense as its a symlink and not a regular file.

site/profile/manifests/ts/hexrot.pp
jhoblitt: That is correct. `'` means no interpolation and `"` means do interpolation.  There is also this other category of strings called "bare words" (I think this term is originally from perl) that are from the early days of puppet.

https://www.puppet.com/docs/puppet/7/lang_data_string.html#lang_data_string_bare_words

cbarria: it was changed to use it as a bare word, thanks!

site/profile/manifests/core/anaconda.pp
cbarria: this was changed to ensure_packages() 👍🏻

dtapiacl (approved):
This looks good to me. Nice work!

I don't have much experience yet with Puppet, but reading through the code, I understood what was going on. Thank you for sharing, Im adding to my bookmarks for future reference.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+class profile::core::anaconda (
+  String $python_env_name,
+  String $python_env_version,
+  Tuple   $conda_packages,
+) {
+  case $facts['os']['family'] {
+    'RedHat': {
+      ensure_packages('wget')
+      exec { 'install_anaconda':
+        command => '/bin/wget https://repo.anaconda.com/archive/Anaconda3-2023.07-2-Linux-x86_64.sh -O /tmp/anaconda_installer.sh && /bin/bash /tmp/anaconda_installer.sh -b -p /opt/anaconda',
```
jhoblitt: It seems like either the URL to the installer or the version should be a class parameter.  E.g. https://github.com/voxpupuli/puppet-python/blob/ece8051906eda06468df17b53125b9e184abab8e/manifests/init.pp#L60

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+        creates => '/opt/anaconda/bin/conda',
+        path    => '/usr/bin:/bin:/opt/anaconda/bin',
+        require => Package['wget'],
+      }
+    }
+    default: {
+      notify { 'unsupported_os':
```
jhoblitt: This could be simplified to `notify { 'This Anaconda install...': }`

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+#   if exist, sets the name of the virtual environment
+#
+# @param python_env_version
+#   if exist, sets the version of python for the virtual environment
+#
+# @param conda_packages
+#   if contains at least 1 tuple (name and channel), will install the package
+#    
+class profile::core::anaconda (
+  String $python_env_name,
```
jhoblitt: It is usually preferable to use `String[1]` as `String` will allow an empty string. E.g. `""`

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+    default: {
+      notify { 'unsupported_os':
+        message => 'This Anaconda installation is only for RedHat family',
+      }
+    }
+  }
+  # Create a Conda environment with Python if parameters are set
+  if ($python_env_name and $python_env_version) {
```
jhoblitt: I don't think this conditional is needed since these parameters are both mandatory.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+#   if contains at least 1 tuple (name and channel), will install the package
+#    
+class profile::core::anaconda (
+  String $python_env_name,
+  String $python_env_version,
+  Tuple   $conda_packages,
+) {
+  case $facts['os']['family'] {
+    'RedHat': {
+      ensure_packages('wget')
```
jhoblitt: This isn't wrong but as a general practice, it is better to script using `curl` since it is part of the default EL package set.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+      creates => "/opt/anaconda/envs/${python_env_name}/",
+      require => Exec['install_anaconda'],
+    }
+    file { '/etc/profile.d/conda_source.sh':
+      ensure  => file,
+      mode    => '0644',
+      require => Exec['create_conda_env'],
+      # lint:ignore:strict_indent
+      content => @("SOURCE"),
+          #!/usr/bin/bash
```
jhoblitt: The shebang isn't needed as it is ignored when a file is sourced.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+          source /opt/anaconda/bin/activate ${python_env_name}
+          | SOURCE
+      # lint:endignore
+    }
+    # Install libmamba inside the environment with conda
+    exec { 'install_libmamba_env':
+      command => "/opt/anaconda/bin/conda install -y -n ${python_env_name} conda-libmamba-solver",
+      creates => "/opt/anaconda/envs/${python_env_name}/lib/libmamba.so",
+      require => Exec['create_conda_env'],
```
jhoblitt: This doesn't seem to be mentioned in the puppet style guide but the community style is to put metaparameters after regular parameters/properties. IOW - `require` should be the last listed param.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
...
+# @param python_env_version
+#   if exist, sets the version of python for the virtual environment
+#
+# @param conda_packages
+#   if contains at least 1 tuple (name and channel), will install the package
+#    
+class profile::core::anaconda (
+  String $python_env_name,
+  String $python_env_version,
+  Tuple   $conda_packages,
```
jhoblitt: Use `Array` unless you are going to specify a list of data types for elements in the array. See https://www.puppet.com/docs/puppet/8/lang_data_abstract#lang_data_abstract_flexible-tuple-data-type

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
+# @summary
+# Sets up repos and symlinks for hexrot
+
+class profile::ts::hexrot {
+  vcsrepo { '/opt/ts_config_mttcs':
+    ensure             => present,
+    provider           => git,
+    source             => 'https://github.com/lsst-ts/ts_config_mttcs.git',
+    keep_local_changes => true,
+    require            => Class[profile::core::anaconda],
```
jhoblitt: Why does this `vcsrepo` resource require conda?

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
...
+        export PYTEST_QT_API="PySide2"
+        export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+        | ENV
+    # lint:endignore
+  }
+  file { '/rubin/mtm2/python':
+    ensure  => directory,
+    owner   => 73006,
+    group   => 73006,
+    require => Class[profile::core::anaconda],
```
jhoblitt: Why does this file resource require conda?

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
...
+    ensure             => present,
+    provider           => git,
+    source             => 'https://github.com/lsst-ts/ts_config_mttcs.git',
+    keep_local_changes => true,
+    require            => Class[profile::core::anaconda],
+  }
+  file { '/etc/profile.d/hexrot_path.sh':
+    ensure  => file,
+    mode    => '0644',
+    require => Vcsrepo['/opt/ts_config_mttcs'],
```
jhoblitt: Metaparameters should be listed last.

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
...
+    # lint:ignore:strict_indent
+    content => @(ENV),
+        #!/usr/bin/bash
```
jhoblitt: The shebang isn't necessary.

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
...
+        export QT_API="PySide2"
+        export PYTEST_QT_API="PySide2"
+        export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+        | ENV
+    # lint:endignore
+  }
+  file { '/rubin/mtm2/python':
+    ensure  => directory,
+    owner   => 73006,
```
jhoblitt: This can be DRY'd up by using a variable.

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,42 @@
...
+    # lint:ignore:strict_indent
+    content => @(ENV),
+        #!/usr/bin/bash
+        export QT_API="PySide2"
+        export PYTEST_QT_API="PySide2"
+        export TS_CONFIG_MTTCS_DIR="/opt/ts_config_mttcs"
+        | ENV
+    # lint:endignore
+  }
+  file { '/rubin/mtm2/python':
```
jhoblitt, line 24: Can we use `/opt/lsst` or `/opt/rubin` instead of creating new top level paths?  `/opt/lsst` is "officially" registered for use by the project.  I should apply for `/opt/rubin` if it is still possible to do so.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,78 @@
+# @summary
+#   Installs anaconda
```
jhoblitt: I still feel that this would be better as an external module instead of as a profile as the functionality isn't "site" specific.

jhoblitt (changes requested):

site/profile/manifests/ts/hexrot.pp
cbarria: I don't think this is movable now since the packages that the team created are already using this /rubin folder =S

site/profile/manifests/core/anaconda.pp
cbarria: yes, this parameter went into a variable 👍🏻

site/profile/manifests/core/anaconda.pp
cbarria: changed the text to the notify 👍🏻

site/profile/manifests/core/anaconda.pp
cbarria: the shebang here is out 👍🏻

site/profile/manifests/core/anaconda.pp
cbarria: I'm moving all the require to last

site/profile/manifests/core/anaconda.pp
cbarria: conditional check was removed 👍🏻

site/profile/manifests/core/anaconda.pp
cbarria: all set to String[1] 👍🏻

site/profile/manifests/ts/hexrot.pp
cbarria: require removed 👍🏻

site/profile/manifests/ts/hexrot.pp
cbarria: also removed

site/profile/manifests/ts/hexrot.pp
cbarria: moved last 👍🏻

site/profile/manifests/ts/hexrot.pp
cbarria: also removed 👍🏻

site/profile/manifests/ts/hexrot.pp
cbarria: this was dried up all 3 into a single one

site/profile/manifests/core/anaconda.pp
cbarria: I change that to be a Hash now

site/profile/manifests/ts/hexrot.pp
cbarria: i even dried up the test also now that I knew how

site/profile/manifests/core/anaconda.pp
cbarria: I use the recommendation from the web page, but this can be changed if it is a problem 👍🏻

site/profile/manifests/core/anaconda.pp
cbarria: I'm investigating how to do it, but I still want to get the other details reviewed, please.

site/profile/manifests/core/anaconda.pp
jhoblitt: `curl` will always be present so using it instead of wget allows the `ensure_packages()` call to be removed completely.

site/profile/manifests/core/anaconda.pp
```diff
@@ -0,0 +1,73 @@
...
+  # Sets libmamba as dependencies solver
+  exec { 'set_libmamba':
+    command => 'conda config --set solver libmamba',
+    unless  => 'conda config --get solver | grep libmamba',
+    path    => ['/usr/bin', '/bin', '/opt/anaconda/bin'],
+    require => Exec['install_libmamba_env'],
+  }
+  #cycle and install conda stuff
+  if !empty($conda_packages) {
+    $conda_packages.each |$package_name, $package_info| {
```
jhoblitt: This is reasonable but I wonder if it would be better to write out a conda env file and then invoke conda only once (per env)?

spec/hosts/roles/hexrot_spec.rb
jhoblitt: Please remove the test for profile classes.  Testing for external classes is OK.

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
 
           %w[
+            anaconda
+            mate
             profile::core::common
             profile::core::debugutils
             profile::core::docker
             profile::core::docker::prune
             profile::core::ni_packages
+            profile::core::x2go
```
jhoblitt: This duplicates `include_examples 'x2go packages', os_facts: os_facts` above.

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
           it { is_expected.to contain_class('docker') }
 
           %w[
+            anaconda
+            mate
             profile::core::common
```
jhoblitt: This duplicates ` include_examples 'common', os_facts: os_facts`

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
             profile::core::debugutils
```
jhoblitt: Replace with `include_examples 'debugutils'`

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
             profile::core::docker
```
jhoblitt: Replace with `include_examples 'docker', docker_version: '24.0.9'`

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
             profile::core::docker::prune
```
jhoblitt: Replace with `it { is_expected.to contain_cron__job('docker_prune') }`

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
             profile::core::ni_packages
```
jhoblitt: There also  isn't a class spec for this one. Please write a shared example for it.

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
           it { is_expected.to contain_class('docker') }
 
           %w[
+            anaconda
```
jhoblitt: This duplicates line below including ` it { is_expected.to contain_class('anaconda').with_anaconda_version('Anaconda3-2023.07-2') }`

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
+            mate
```
jhoblitt: This duplicates `          it { is_expected.to contain_class('mate') }`

site/profile/manifests/ts/hexrot.pp
```diff
@@ -0,0 +1,44 @@
+# @summary
+# Sets up repos and symlinks for hexrot
+
+class profile::ts::hexrot {
+  vcsrepo { '/opt/ts_config_mttcs':
```
jhoblitt, line 5: This is an important resource which needs spec coverage.

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -32,16 +32,85 @@
...
             profile::core::common
             profile::core::debugutils
             profile::core::docker
             profile::core::docker::prune
             profile::core::ni_packages
+            profile::core::x2go
+            profile::ts::hexrot
+            profile::ts::nexusctio
```
jhoblitt: Please write a shared example for `            profile::ts::nexusctio`

jhoblitt (changes requested):
This generally looks good.  I would like to see some improvements to the spec coverage, as there are some sins from the past, even though this is bordering on being beyond the scope of this PR.

spec/hosts/roles/hexrot_spec.rb
cbarria: removed 👍🏻

spec/hosts/roles/hexrot_spec.rb
cbarria: removed 👍🏻

spec/hosts/roles/hexrot_spec.rb
cbarria: replaced and removed 👍🏻

site/profile/manifests/ts/hexrot.pp
cbarria: yes, this was missing and was added just now

spec/hosts/roles/hexrot_spec.rb
cbarria: changed

spec/hosts/roles/hexrot_spec.rb
cbarria: removed/changed 👍🏻

spec/hosts/roles/hexrot_spec.rb
cbarria: i couldn't use the shared_examples for docker because the hexrot uses `storage_driver: 'devicemapper'`
but i included a test for this condition.

spec/hosts/roles/hexrot_spec.rb
cbarria: great idea, just test the resource, not the class 👍🏻  changed.

spec/hosts/roles/hexrot_spec.rb
cbarria: removed the self class test

spec/hosts/roles/hexrot_spec.rb
cbarria: support/spec created for ni_packages 👍🏻

spec/hosts/roles/hexrot_spec.rb
cbarria: support/spec created for nexusctio 👍🏻

spec/hosts/roles/hexrot_spec.rb
```diff
@@ -6,6 +6,8 @@
 
 describe "#{role} role" do
   on_supported_os.each do |os, os_facts|
+    next if os =~ %r{almalinux-8-x86_64}
```
jhoblitt: Why is hexrot using EL8?

cbarria: sorry, silly me, let me fix this right now...

jhoblitt (approved):
LGTM.

PR #933, opened 2023-09-29
(role/hypervisor) fix libvirt startup on EL9

jhoblitt:
libvirt startup seems to have regressed again on EL9 on newly provisioned nodes. Hopefully, pedantically enumerating the socket units will be a durable fix.


cbarria (approved):
Looks Great 👍🏻

PR #937, opened 2023-10-02
(common) enable sssd startup of services; disable socket activation

jhoblitt:
`bodgit/sssd` was enabling systemd socket unit activation for sssd sub-services while the puppet/sssd module does not.  This created a problem as the `sssd.conf` was setup of socket activation.  As we were unable to get sssd socket activation to work as expeected on EL9.2 (odd).  Thus, we are reverting to sssd driven activation of sssd sub-services, which is consitent with the configuration created by `ipa-client-install`.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -341,7 +341,11 @@ sssd::main_config:
   sssd:
     config_file_version: 2
     domains: "lsst.cloud"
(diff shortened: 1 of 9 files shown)


cbarria (approved):
Looks Great 👍🏻  thanks for fixing it

PR #942, opened 2023-10-04
(role/tang) fwv

jhoblitt:
Based on:
- #974
- #978
Related to https://github.com/lsst-it/foreman_templates/pull/11


cbarria (approved):
Looks Great 👍🏻

hieradata/site/dev/role/tang.yaml
```diff
@@ -0,0 +1,9 @@
+---
+profile::core::firewall::firewall:
+  "200 accept tang":
+    proto: "tcp"
+    state: "NEW"
+    ipset: "dev src"
+    dport: "7500"
+    action: "accept"
+    require: "Ipset::Set[dev]"
```
bastelfreak: At a previous ISP we had a collector that ensured all ipsets are configured before the firewall rules. With that single line of code you could get rid of all those `require` statements.

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'tang01.dev.lsst.org', :sitepp do
+  on_supported_os.each do |os, facts|
```
bastelfreak: It gets a bit confusing with this `facts` as local variable and the rspec specific `facts` in line 10. At Vox Pupuli we switched to:
```suggestion
  on_supported_os.each do |os, os_facts|
```

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    next unless os =~ %r{almalinux-9-x86_64}
+
+    context "on #{os}" do
+      let(:facts) do
+        override_facts(facts,
```
bastelfreak:
```suggestion
        override_facts(os_facts,
```

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        {
+          role: 'tang',
+          site: 'dev',
+        }
+      end
+
+      it { is_expected.to compile.with_all_deps }
+
+      include_examples 'vm'
+      include_context 'with nm interface'
```
bastelfreak, line 31: Nice, that's rarely used/known.

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      it { is_expected.to have_nm__connection_resource_count(1) }
+
+      context 'with enp1s0' do
+        let(:interface) { 'enp1s0' }
+
+        it_behaves_like 'nm enabled interface'
+        it_behaves_like 'nm ethernet interface'
```
bastelfreak, line 38: nice again, also rarely used.

hieradata/site/dev/role/tang.yaml
jhoblitt: That is a very good point... I have no idea why I didn't do that... I will update it.

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
jhoblitt: Ya, that has been driving me crazy too...

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
jhoblitt: I need to write a rspec extension gem for testing that puppet mod... some day I will find the time...

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
jhoblitt: It is just an alias for `include_examples`.  I should probably be consistent and use only one of the aliases instead of both.

jhoblitt: although I do like the extra doc output from `it_behaves_like`...

spec/support/spec/ipset.rb
```diff
@@ -36,4 +36,24 @@
       ],
     )
   end
+
+  it do
+    is_expected.to contain_ipset__set('dev').with_set(
+      %w[
```
bastelfreak, line 42: I had a discussion recently with another dev about unit tests. There isn't much logic around for the ipsets. You define it in hiera and the resources will be created like that. Is it worth it writing tests for that? I don't know. I often write them to feeld good, but I don't think they provide much value. Another idea that might bring more value: count how many ipset resources you have in the catalog (like documented here: https://rspec-puppet.com/documentation/classes/#testing-the-number-of-resources-of-a-specific-type).

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
bastelfreak: If you ask Ewoud about it he can probably explain the differences and when to use include_examples vs it_behaves_like, but that's beyond my knowledge :D

spec/support/spec/ipset.rb
jhoblitt: So the reason I do that is basically double entry accounting.  I don't cut'n'paste the values from hiera but instead re-key them or cut'n'paste from the source of truth as a way of making sure the hiera value is actually correct.  It has also been helpful during refactors to catch when a class is accidentally dropped from the manifest. And it makes me feel better too... I would very much prefer to do this in rspec-beaker...

bastelfreak: yes. beaker would be cool. But this is a good check to catch wrong typos in the data.

spec/hosts/nodes/tang01.dev.lsst.org_spec.rb
jhoblitt: Uh, thanks...

```
 182 files changed, 520 insertions(+), 530 deletions(-)
```

tuxmea (approved):
LGTM

PR #951, opened 2023-10-06
(site/ls/role/foreman) add IT-CORE-SERVICES|IT-HYPERVISOR|IT-BMC pools

jhoblitt:


julioconstanzo (changes requested):
Range on IT-BMC has typo. Last IP from that range should be 142.250 (251 and 252 are SVI)

julioconstanzo (approved):
Looks good.

PR #952, opened 2023-10-06
(site/cp/role/foreman) fix IT-CORE-SERVICES|IT-HYPERVISOR|IT-BMC pools

jhoblitt:
The highest address in the dhcp pools overlapped with SVIs.


julioconstanzo (approved):
looks good.

PR #955, opened 2023-10-14
new ccs releases for ccs clusters

glennmorris:
Not to be merged until CCS releases are ready.


tony-johnson (approved):
We have tested this on auxtel at BTS. The CCS release is now finalized, so we should proceed to merge this into production.

PR #956, opened 2023-10-16
(cluster/*ccs) rm disabling of sudo config purging

jhoblitt:
This is no longer necessary and is definitely not desirable.

### DIFF
--- a/hieradata/cluster/auxtel-ccs.yaml
+++ b/hieradata/cluster/auxtel-ccs.yaml
@@ -17,8 +17,6 @@ postfix::configs:
     value: "hash:/etc/postfix/sasl_passwd"
   smtp_sasl_mechanism_filter:
     value: "plain, login"
-
-sudo::purge: false  # FIXME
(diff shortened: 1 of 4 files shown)


glennmorris:
So are there no non-puppetized sudoers files left?
The only one I can think of is /etc/sudoers.d/x2goserver.
lsst-control/site/profile/manifests/core/x2go.pp refers to it, but only the permissions, not the content?
Is that enough to stop it being removed?

jhoblitt:
> So are there no non-puppetized sudoers files left? The only one I can think of is /etc/sudoers.d/x2goserver. lsst-control/site/profile/manifests/core/x2go.pp refers to it, but only the permissions, not the content? Is that enough to stop it being removed?

That is correct, https://github.com/lsst-it/lsst-control/blob/production/site/profile/manifests/core/x2go.pp#L15-L21 will prevent the x2go config from being purged.

glennmorris (approved):

PR #957, opened 2023-10-18
(.puppet-lint.rc) disable legacy_facts check

jhoblitt:
This will need to be re-enabled/resolved in order to migrate to puppet8.


dtapiacl (approved):
Great thanks for cleaning that up.

dtapiacl (approved):

PR #959, opened 2023-10-19
(Puppetfile) replace ghoneycutt/hosts with puppet/hosts

jhoblitt:
The fork of ghoneycutt/hosts that was in use was not compatible with stdlib 9.x.  It was less effort to write a new module from scratch rather than updating the fork.

See: https://github.com/voxpupuli/puppet-hosts


dtapiacl (approved):
LGTM.

PR #960, opened 2023-10-20
 (common) fix sshd_config for csilva_b user

jhoblitt:


dtapiacl (approved):
LGTM. I noticed i was approaching my unit test in the incorrect manner, this looks a lot cleaner.

PR #964, opened 2023-11-01
(node/foreman.*) migrate dhcp pool configuration from site/role -> node layer(s)

jhoblitt:
This is to facilitate spinning up test VMs with the foreman role that do not become a rogue dhcp server.


dtapiacl (approved):
LGTM.

PR #971, opened 2023-11-08
(hiera) use %{facts.os.*} instead of %{os.*}

jhoblitt:
This is to avoid a catalog compilation failure when a class uses `$os` as a parameter.


cbarria (approved):
Looks Great 👍🏻

PR #972, opened 2023-11-10
(*) Capitalize bare words used to ref class names

jhoblitt:
Bare string handling in puppet is bazaar...

Resolves these errors:

```
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Invalid relationship: File[/etc/sudoers.d/perfsonar_sudo] { require => Class[perfsonar] }, because Class[perfsonar] doesn't seem to be in the catalog
```

```
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Invalid relationship: File[/etc/NetworkManager/dispatcher.d/50-eno1] { notify => Class[network] }, because Class[network] doesn't seem to be in the catalog
```

### DIFF
--- a/hieradata/role/perfsonar.yaml
+++ b/hieradata/role/perfsonar.yaml
@@ -56,7 +56,7 @@ files:
   # perfsonar packaging installs this one file with the wrong mode
   "/etc/sudoers.d/perfsonar_sudo":
     mode: "0440"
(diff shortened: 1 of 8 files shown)


cbarria (approved):
nice catch, Looks Great! 👍🏻

PR #977, opened 2023-11-14
(node/auxtel-control01) remove network config

dtapiacl:
Clean up of network configuration of now retired auxtel-control-01.cp.lsst.org.


jhoblitt (approved):
RIP

PR #978, opened 2023-11-14
(common) prepare to use restic in multiple roles

jhoblitt:
https://github.com/lsst-it/lsst-puppet-hiera-private/pull/104


cbarria (approved):
Looks Great 👍🏻

PR #979, opened 2023-11-15
(auxtel-hcu02.cp) Summit Deployment

dtapiacl:
auxtel-hcu02.cp.lsst.org  deployment at the summit.

### DIFF
--- a/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
+++ b/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
@@ -63,3 +63,5 @@ nm::connections:
 
       [ipv6]
       method=disabled
+
+ccs_hcu::imanager: true
--- a/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
+++ b/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
@@ -1,0 +1,65 @@
+---
+nm::connections:
+  eno1:
+    content: |
+      [connection]
+      id=eno1
+      uuid=250c1a0b-95c3-4798-a4df-d864e60d3b63
+      type=ethernet
+      interface-name=eno1
+
+      [ethernet]
+
+      [ipv4]
+      method=auto
+
+      [ipv6]
+      method=disabled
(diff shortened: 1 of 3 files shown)


hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
```diff
@@ -0,0 +1,67 @@
...
+      [ethernet]
+
+      [ipv4]
+      method=manual
+      address1=192.168.1.1/24
+
+      [ipv6]
+      method=disabled
+
+ccs_hcu::imanager: true
```
jhoblitt, line 67: Please check with @tony-johnson and/or @glennmorris about this key -- I am unsure if it should be on both HCUs.

hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
```diff
@@ -0,0 +1,67 @@
+---
+nm::connections:
```
jhoblitt, line 2: This is OK as-is but please switch to using the yaml style config instead of heredoc'd strings in the future.  I have a script to convert this style to the pure yaml style.

jhoblitt (changes requested):
This generally looks good. Please check with Tony/Glenn as to the ccs config key. The format of the PR title also needs to be fixed.

hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
glennmorris: It was me who added that commit. It should really be in the atschu role (per the commit message), but I did not want to mess up the existing hand-installed hcu01 at this stage.

glennmorris (approved):
Fine be my (modulo Josh's comments re yaml style).

hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
dtapiacl: Copy that.

jhoblitt (approved):
LGTM.

PR #982, opened 2023-11-16
(profile::core::gpio) fwv

jhoblitt:


glennmorris (commented):
I guess this is fine, assuming it will be a no-op on RHEL7?

jhoblitt:
> I guess this is fine, assuming it will be a no-op on RHEL7?

Correct. There will be no char device to match the name pattern, so nothing will happen.

glennmorris (dismissed):

glennmorris:
Actually, now that I think about it, shouldn't this be part of the [ccs_hcu](https://github.com/lsst-it/puppet-ccs_hcu) module?

jhoblitt:
@glennmorris I feel like this is really in role/profile territory.  We are going to use the same udev rules for rpis so I think I'm going to turn this into a profile class.

jhoblitt:
This has been reworked to add a new `profile::core::gpio` profile.

glennmorris (approved):
Looks great, thank you!

PR #986, opened 2023-11-16
(role/rpi) add profile::core::docker

jhoblitt:


cbarria (approved):
Doesnt look great, but it will

PR #995, opened 2023-11-17
(profile::pi::rubinhat) fwv 

jhoblitt:


wvreeven (approved):
Looks good to me.

spec/support/spec/rubinhab.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
fritzm: Typo on file name?

fritzm (approved):
One typo'd filename?  Otherwise LGTM.

spec/support/spec/rubinhab.rb
jhoblitt: Fixed.  The name of the file actually isn't significant as every file under `spec/support/spec` is loaded.

PR #996, opened 2023-11-17
(profile::pi::ftdi) fwv

jhoblitt:


fritzm (approved):
LGTM

wvreeven (commented):
I tested the rule locally with a serial-to-USB converter cable as used at the summit and the group permission was set as expected. Looks good to me.

PR #998, opened 2023-11-20
(auxtel-hcu02.cp) OS Redeployment

dtapiacl:
This pull request updates the network configuration and tests for Centos 7 redeployment of auxtel-hcu02.cp.lsst.org.

### DIFF
--- a/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
+++ b/hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
@@ -1,67 +1,23 @@
 ---
-nm::connections:
-  eno1:
-    content: |
-      [connection]
-      id=eno1
-      uuid=250c1a0b-95c3-4798-a4df-d864e60d3b63
-      type=ethernet
-      interface-name=eno1
-
-      [ethernet]
-
-      [ipv4]
-      method=auto
-
-      [ipv6]
-      method=disabled
-  enp8s0:
-    content: |
-      [connection]
-      id=enp8s0
-      uuid=de9904c8-9577-1a17-36b1-34b94132f06a
-      type=ethernet
-      autoconnect=false
-      interface-name=enp8s0
-
-      [ethernet]
-
-      [ipv4]
-      method=disabled
-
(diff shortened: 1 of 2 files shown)


spec/hosts/nodes/auxtel-hcu02.cp.lsst.org_spec.rb
```diff
@@ -28,37 +28,41 @@
       it { is_expected.to compile.with_all_deps }
 
       include_examples 'baremetal no bmc'
-      include_context 'with nm interface'
-      it { is_expected.to have_nm__connection_resource_count(4) }
```
jhoblitt, line 32: Having a resource count for `network__interface` is a good idea to make sure extra interfaces aren't being created by accident.

jhoblitt (changes requested):
A resource count check would be a good ideal. Otherwise, this looks good.

hieradata/node/auxtel-hcu02.cp.lsst.org.yaml
```diff
@@ -1,67 +1,21 @@
...
-      [ethernet]
-
-      [ipv4]
-      method=manual
-      address1=192.168.1.1/24
-
-      [ipv6]
-      method=disabled
-
-ccs_hcu::imanager: true
```
glennmorris, line 67: Why are you removing the ccs_hcu::imanager setting?
I would like it to stay.

jhoblitt: I missed that it was being removed.  Is there any reason we can't move this to the role?

glennmorris: Long term it should be in the role, yes.
But I don't want to affect hcu01 at the moment, where it seems imanager was installed by hand ages ago.

glennmorris (approved):
Looks good to me now, thanks.

cbarria (approved):
Looks Great 👍🏻

jhoblitt (approved):
LGTM

PR #1002, opened 2023-11-21
(trewa0[123].cp) remove trewa entries

dtapiacl:
Clean up of Trewa cluster.


jhoblitt (commented):
I think the `ts-csc` role can probably be removed as well.  I think only the `trewa` nodes were using it but I would double check it isn't being used in BTS. https://foreman.cp.lsst.org/hosts?search=params.role+%3D+ts-csc&page=1

dtapiacl:
> I think the `ts-csc` role can probably be removed as well. I think only the `trewa` nodes were using it but I would double check it isn't being used in BTS. https://foreman.cp.lsst.org/hosts?search=params.role+%3D+ts-csc&page=1

I checked foreman.ls and i dont see any other nodes consuming that role. I will clean up the role entries as well as there are no clusters using it.

jhoblitt (approved):
This is a perfect PR as it only deletes lines. Nicely done.

PR #1003, opened 2023-11-21
(common) pin el7/aarch64 to puppet agent 7.24.0

jhoblitt:
As 7.24.0 is the latest version of puppet that will be officially packaged for el7/aarch64.

See http://yum.puppet.com/puppet7/el/7/aarch64/


wvreeven (approved):
lgtm

PR #1004, opened 2023-11-21
(role/{ess,laserrpi}) rm uneeded docker keys

jhoblitt:
These values are currently hard coded in `profile::core::docker` and have no effect.


wvreeven (approved):
lgtm

PR #1010, opened 2023-11-27
(foreman.cp) add dhcp pool vlan1609 TMA-Network

julioconstanzo:
A new DHCP pool was created for VLAN1609 (TMA Network). Test and IPAM were updated to reflect these changes.

The subnet already existed so this is to create a DHCP pool far away from the already static IPs on the TMA network. Also, this is to create the FQDNs of all existent devices inside that subsystem.

### DIFF
--- a/hieradata/node/foreman.cp.lsst.org.yaml
+++ b/hieradata/node/foreman.cp.lsst.org.yaml
@@ -123,6 +123,13 @@ dhcp::pools:
     range:
       - "139.229.170.64 139.229.170.191"  # ~/25
     search_domains: "%{alias('dhcp::dnsdomain')}"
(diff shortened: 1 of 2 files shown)


jhoblitt (approved):
LGTM. Thank you for reserving the dhcp pool to ipam.  It would probably be a good idea to reserve .251 & .252 to ipam as well to prevent them from being auto-allocated in the future.

julioconstanzo:
> LGTM. Thank you for reserving the dhcp pool to ipam. It would probably be a good idea to reserve .251 & .252 to ipam as well to prevent them from being auto-allocated in the future.

.251 and .252 allocated in IPAM. Thanks.

PR #1014, opened 2023-11-30
(vpn{1-2}.ls) Network configuration and tests

dtapiacl:
This PR adds network configuration and tests to vpn1.ls.lsst.org and vpn2.ls.lsst.org.


cbarria (approved):
Looks Great 👍🏻

PR #1018, opened 2023-11-30
(cp/role/nfsserver) add rsphome share for nublado

cbarria:
A new NFS was requested to replace jhome on the new instance of Nublado for the summit.

### DIFF
--- a/hieradata/site/cp/role/nfsserver.yaml
+++ b/hieradata/site/cp/role/nfsserver.yaml
@@ -77,6 +77,13 @@ nfs::nfs_exports_global:
       139.229.165.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.170.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.191.0/25(rw,nohide,insecure,no_subtree_check,async,root_squash)
+  /data/rsphome:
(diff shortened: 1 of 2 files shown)


hieradata/site/cp/role/nfsserver.yaml
```diff
@@ -77,6 +77,13 @@ nfs::nfs_exports_global:
       139.229.165.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.170.0/24(rw,nohide,insecure,no_subtree_check,async,root_squash)
       139.229.191.0/25(rw,nohide,insecure,no_subtree_check,async,root_squash)
+  /data/rsphome:
```
jhoblitt, line 80: Is this going to be migrated to the new storage cluster or should it be hosted on yagan?  I don't think we have a clear plan for what we are going to do with nfs exports.

cbarria: I created the other 2 rsphomes (TTS and BTS) in the same place the jhome was... I think jhome is retiring but cant be cleaned yet, they need to backup the homes of the users first.

csilva-cl (approved):
code looks fine.

granted we still need to solve the "nfs" problem.

PR #1020, opened 2023-12-06
(role/{atshcu,ccs-hcu}) add telnet package

glennmorris:
This is for basic communication with attached hardware devices.


dtapiacl (approved):
Looks good to me.

PR #1022, opened 2023-12-13
Add stationary /dev/serial* symlinks

fritzm:
Adds to existing udev script for rubin pi units -- creates `/dev/serial*` symlinks that are linked to actual hardware resources on the Broadcom chip.  We can use these in higher level software instead of `/dev/ttyAMA*`, and be sure they won't move around between kernel releases.


wvreeven (approved):
Looks good to me.

cbarria (approved):
Looks Great!

PR #1025, opened 2023-12-21
(foreman.cp) add public dns to guest wifi

csilva-cl:
add public dns to guest wifi at the summit.


spec/hosts/nodes/foreman.cp.lsst.org_spec.rb
dtapiacl, line 1: I think the catalog is expecting the following order of, nameservers: ['1.0.0.1', '1.1.1.1', '8.8.8.8'] . As the rake is failing, other than this it looks good to me.

csilva-cl: the spec is expecting the order of the nameservers in the config ['1.1.1.1', '1.0.0.1', '8.8.8.8'], I fixed it in the last push

dtapiacl: I see, it looks good.

dtapiacl (approved):
LGTM.

PR #1029, opened 2024-01-10
(role/dnscache) update dns forwarder ip

dtapiacl:
This PR is to update the DNS forwarder from Crux to DC-LS-2.

### DIFF
--- a/hieradata/role/dnscache.yaml
+++ b/hieradata/role/dnscache.yaml
@@ -42,10 +42,4 @@ dns::zones:
     zonetype: "forward"
     forward: "only"
     forwarders:
-      - &crux "139.229.136.23"
-  "136.229.139.in-addr.arpa":
-    reverse: true
-    zonetype: "forward"
-    forward: "only"
-    forwarders:
-      - *crux
+      - &dc-ls-2 "139.229.135.58"


hieradata/role/dnscache.yaml
```diff
@@ -42,10 +42,10 @@ dns::zones:
     zonetype: "forward"
     forward: "only"
     forwarders:
-      - &crux "139.229.136.23"
-  "136.229.139.in-addr.arpa":
+      - &dc-ls-2 "139.229.135.58"
+  "135.229.139.in-addr.arpa":
```
jhoblitt: I don't think 139.229.135/24 is correct.

jhoblitt (changes requested):

hieradata/role/dnscache.yaml
jhoblitt: I still think the reverse zone is incorrect.

jhoblitt (changes requested):
Please fix the reverse zone.

hieradata/role/dnscache.yaml
```diff
@@ -42,10 +42,10 @@ dns::zones:
     zonetype: "forward"
     forward: "only"
     forwarders:
-      - &crux "139.229.136.23"
-  "136.229.139.in-addr.arpa":
+      - &dc-ls-2 "139.229.135.58"
+  "58.135.229.139.in-addr.arpa":
```
jhoblitt: This is still incorrect.  139.229.135/24 has IT chile servers in it.  We can not delegate this subnet to the PMO.

jhoblitt (changes requested):
Please delegate the correct subnet.

hieradata/role/dnscache.yaml
csilva-cl: I created the pmo records in r53. The delegation can be removed.

jhoblitt (approved):
Looks reasonable.

PR #1030, opened 2024-01-10
(node/auxtel-archiver.cp) configuration change to VM

cbarria:
Removed the old network config for Centos 7 and changed the test to match the new VM.


spec/hosts/nodes/auxtel-archiver.cp.lsst.org_spec.rb
```diff
@@ -29,72 +29,6 @@
 
       it { is_expected.to compile.with_all_deps }
 
-      include_examples 'baremetal'
```
jhoblitt, line 32: Could you include the `vm` examples?

jhoblitt (changes requested):

spec/hosts/nodes/auxtel-archiver.cp.lsst.org_spec.rb
cbarria: had to open the shared examples because the active_profile on the VMs in the vcenter is set to "balanced"

jhoblitt: That is a bug we should fix. https://github.com/lsst-it/lsst-control/blob/production/hieradata/common/virtual/kvm.yaml needs to be copied to be `vmware.yaml` and that should set the expected tuned profile.

cbarria: this was added to the virtual folder and Test was changed back to 'vm'

jhoblitt (approved):
LGTM.

PR #1031, opened 2024-01-12
(WIP) VPN deployment using puppet

dtapiacl:
This is PR was created to deploy the vpn service at Rubin Observatory.

Ticket of reference: IT-4981


site/profile/manifests/core/openvpnas.pp
```diff
@@ -0,0 +1,24 @@
+class profile::core::openvpnas {
```
jhoblitt, line 1: There are several openvpn modules on the forge.  Is there a reason why we can not use one of the existing modules?

dtapiacl: Hi @jhoblitt. The reason for this is that since it was my first puppet installation, I really wanted to learn how to automate this whole process using a more traditional approach with puppet.  This process helped me learn and understand class creations and roles more in-depth.

jhoblitt (commented):
What is the plan for configuring openvpn?

dtapiacl:
> What is the plan for configuring openvpn?

Hey @jhoblitt, the VPN service was already previously configured by @hstockeb. I imported the database configuration onto this deployment to get it up and running. @hstockeb could provide more information for its specific configuration and setup.

jhoblitt:
@dtapiacl Multiple openvpn modules on the puppet forge support configuration.  It would be a good idea to automate as much of the configuration as possible as a DR measure.  Installing the packages is a good first step but the configuration is much more valuable.

jhoblitt:
@dtapiacl To be pedantic, I see 3 good courses of action. 1) use an existing forge module to configure openvpn. 2) modify/fix an existing forge module to configure openvpn. 3) write a new module to manage openvpn and publish it the forge.  Option #3 should only be a last resort after attempting option 1/2.

It is not a good idea to add profile(s) to this the control repo that duplicate the functionality of existing forge modules.  Every line of code has a maintenance burden.  As a general rule, we should try to add the minimum on-going maintenance requirement as possible onto our fellow engineers.

jhoblitt (changes requested):
I would like to see configuration of the service and to use a forge module. E.g. https://forge.puppet.com/modules/puppet/openvpn/readme

PR #1032, opened 2024-01-16
(common) rm telegraf from all hosts

jhoblitt:
This is first step in migrating away from telegraf that uninstalls telegraf from all hosts.  The second step will be to rm the telegraf puppet module and supporting code from the control repo.


jhoblitt:
The librarian-puppet failure is bogus.  It has failed on multiple branches this morning.

cbarria (approved):
Looks Great 👍🏻

PR #1033, opened 2024-01-16
(node/comcam-hcu03) set imanager true

glennmorris:
It seems that imanager had been installed by hand on comcam-hcu03, and for some reason stopped working recently.
Letting puppet manage imanager fixes this.


glennmorris:
PS after this is merged, will need to update cp_production before moving comcam-hcu03 off the IT-5030/ccs-comcam-hcu03-gpio branch.

jhoblitt:
The librarian-puppet failure seems to be something to do the forge API so I will force merge this.

jhoblitt (commented):
LGTM.

PR #1034, opened 2024-01-16
(role/logictimer) fwv

jhoblitt:
Per discussion via slack, the `laserrpi` role does not need gpshat support but a new role for the "logic timer" pis does.


fritzm (approved):
LGTM -- thanks!

jhoblitt:
The librarian-puppet failure is bogus. Merging.

PR #1035, opened 2024-01-29
(cluster/manke) add vlan2504 interface

cbarria:
to communicate the k8s pods from the HTCondor Cluster to the Scheduler we need to set the VLAN in the trunk interface

### DIFF
--- a/hieradata/cluster/manke.yaml
+++ b/hieradata/cluster/manke.yaml
@@ -1,166 +1,167 @@
 ---
 clustershell::groupmembers:
-  manke: {group: "manke", member: "manke[01-10]"}
+  manke:
+    group: "manke"
+    member: "manke[01-10]"
 profile::core::ospl::enable_rundir: true
 profile::core::rke::enable_dhcp: true
 nm::connections:
   eno1np0:
-    content: |
-      [connection]
-      id=eno1np0
-      uuid=f330f829-20cc-b829-67b0-18086a5fe6fa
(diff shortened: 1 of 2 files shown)


jhoblitt (commented):
I have been converting the heredoc style to nested hash style when I touch files.  This script should get you most of the way there but the changes need to be carefully checked as it will quote some things that should be bare words.

```ruby
#!/usr/bin/ruby

require 'psych'
require 'yaml'
require 'iniparse'

filename = ARGV[0]

# convert nm content from string -> hash
yaml = YAML.load_file(filename)
yaml['nm::connections'].each do |iface, config|
  ini = IniParse.parse(config['content'])
  config['content'] = ini.to_h
end

# https://github.com/ruby/psych/issues/322#issuecomment-328408276
ast = Psych.parse_stream(YAML.dump(yaml))

# First pass, quote everything
ast.grep(Psych::Nodes::Scalar).each do |node|
  node.plain  = false
  node.quoted = true
  node.style  = Psych::Nodes::Scalar::DOUBLE_QUOTED
end

ast.grep(Psych::Nodes::Mapping).each do |node|
  node.children.each_slice(2) do |k, _|
    k.plain  = true
    k.quoted = false
    k.style  = Psych::Nodes::Scalar::ANY
  end
end

File.write(filename, ast.yaml)
```

jhoblitt (approved):
LGTM. Thank you for converting to the hash format.

PR #1036, opened 2024-01-29
(role/atshcu) enable imanager

glennmorris:
The imanager module should be enabled for the atshcu role, but to date has not been.
This was to avoid modifying the existing hand-written config on auxtel-hcu01.
However, the existing manual configuration randomly stops working at times (on reboot?), so the time has come to let puppet manage it.

To be safe, this should not be deployed during an auxtel run (ie not this week).
Maybe next week should be ok? When the time comes, cp_production would need an update.


jhoblitt (approved):
This looks good.  If the rollout needs to wait for a maint window, it is probably best not to merge it to production until the window opens.

tony-johnson (approved):
Looks good to me.

PR #1038, opened 2024-01-30
(site/ls/cluster/lsstcam-ccs) set DAQ version to R5-V8.1

glennmorris:


tony-johnson (approved):
Thanks. Looks good to me.

PR #1039, opened 2024-01-31
It 5068 daq v8.2

jgthayer:
In the end, there was a bug and V8.1 became V8.2. 

The puppet changes that were added in [IT-4661] mean that this can be deployed to production and the DAQ release won't be switched to automatically, so full speed ahead.


jhoblitt (approved):
LGTM.  Needs to be rebased before merge.

PR #1040, opened 2024-02-01
(gis-bastion01.cp) EL9 redeployment of GIS-Bastion01

dtapiacl:
This PR is to redeploy gis-bastion01.cp.lsst.org with Alma Linux 9 as part of the OS migration process.


jhoblitt (approved):
LGTM

PR #1043, opened 2024-02-02
(ccs) set daq client version to R5-V8.1

glennmorris:


tony-johnson (approved):
Looks good to me. When we switch to this new version (by changing the symlink) we will need to coordinate with Gregg as he needs to restart the DAQ into the new release.

PR #1045, opened 2024-02-09
(node/comcam-mcm.{cp,tu}) add lockmanager service

glennmorris:
It seems that this service was previously added by hand to both systems.

Ideally the service specification would move to the ccs-mcm role, but at present the services are not unified across all hosts having that role.

### DIFF
--- a/hieradata/node/comcam-mcm.cp.lsst.org.yaml
+++ b/hieradata/node/comcam-mcm.cp.lsst.org.yaml
@@ -44,6 +44,7 @@ ccs_software::services:
     - "mmm"
     - "comcam-mmm"
     - "cluster-monitor"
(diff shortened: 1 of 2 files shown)


jhoblitt (commented):
Does it make sense to migrate the services to a role layer?

glennmorris:
> Does it make sense to migrate the services to a role layer?

:smiley: 

In theory yes, but in practice it is in't straightforward at present, so I would like to defer that for another time:

> Ideally the service specification would move to the ccs-mcm role, but at present the services are not unified across all hosts having that role.

jhoblitt (approved):
This is fine but I'd really like to have a plan in place to sort out of the roles.

PR #1046, opened 2024-02-09
Make networkmanager optional

bastelfreak:

PR #1047, opened 2024-02-09
Implement foreman test environment

bastelfreak:
I added an INSTALL.md with some instructions. This doesn't pass yet. So far this works:

```
puppet agent -t accounts,prometheus,chrony,yumrepo,auditd,tftp,convenience,debugutils,rsyslog,discovery,puppetserver,host,irqbalance,ssh,lldpd,sysstat
```

Somehow something breaks the dynflow database if I apply the whole catalog and then foreman doesn't start anymore. I will investigate next week.

### DIFF
--- a/hieradata/role/foreman.yaml
+++ b/hieradata/role/foreman.yaml
@@ -203,8 +203,9 @@ profile::core::foreman::foreman_config:
   bmc_credentials_accessible: {value: false}  # disable bmc pass in enc yaml
   default_pxe_item_global: {value: "discovery"}
   destroy_vm_on_host_delete: {value: true}
-  discovery_fact_column: {value: "ipmi_ipaddress,ipmi_macaddress"}
-  discovery_hostname: {value: "ipmi_macaddress,discovery_bootif"}
+  # on older foreman/puppet-foreman it wasn't idempotent to set this, but it works withforeman 3.8/ puppet-foreman 24.1.0
+  discovery_fact_column: {value: '["ipmi_ipaddress","ipmi_macaddress"]'}
+  discovery_hostname: {value: '["ipmi_macaddress","discovery_bootif"]'}
   entries_per_page: {value: 100}
   # remove "docker*" from default excluded_facts
   # XXX using block scalar style results in the double quotes being preceeded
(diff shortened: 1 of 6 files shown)


bastelfreak:
@jhoblitt this works for me now:

```
[root@lsst ~]# puppet agent -t
Info: Using environment 'bastelfreak'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Warning: Facter: invalid or malformed lldp data: MalformedDataError: no lldp interface data found
Warning: The current total number of facts: 2490 exceeds the number of facts limit: 2048
Notice: Requesting catalog from $host:8140 ($ip)
Notice: Catalog compiled by $host
Info: Caching catalog for $host
Info: Applying configuration version '1707679430'
Notice: System Reboot Required Because:
  audit => modified
Notice: Applied catalog in 24.96 seconds
[root@lsst ~]#
```

There were a few minor hiera changes and I had to update the theforeman/foreman module. I only tested the latest version 24.1.0. The majority of the changes are related to the apache in front of foreman, not foreman itself.

I also think that the foreman module has some ordering issues and cannot apply the foreman_config_entry during the first puppet run and a bunch of foreman related packages are installed afterwards. I didn't dig too deep into this. They were all applied without issues during the second run and the third one is idempotent and has no changes.

To make the testing a bit easier, I made a few of your profiles optional/opt-out and provided those changes as dedicated PRs:

* #1048 
* #1046 
* #1049

hieradata/role/foreman.yaml
```diff
@@ -81,6 +81,7 @@ foreman::oauth_active: true
 #foreman::oauth_consumer_key:  # secret
 #foreman::oauth_consumer_secret:  # secret
 foreman::passenger: false  # use puma; param removed in theforman/foreman >= 17.0.0
+# we need to figure out how to configure columns in newer foreman versions
 foreman::plugin::column_view::columns:
```
bastelfreak, line 85: the plugin doesn't exist anymore because the code moved into foreman core. The Hiera data is currently unused. I didn't investigate yet how to configure the columns on modern foreman.

bastelfreak:
For what it's worth, an upgrade to 3.9.1 works as well:

```
$ git diff bastelfreak -- hieradata/site/test.yaml
diff --git a/hieradata/site/test.yaml b/hieradata/site/test.yaml
index 9c8d5879..fe389676 100644
--- a/hieradata/site/test.yaml
+++ b/hieradata/site/test.yaml
@@ -28,8 +28,8 @@ profile::core::yum::versionlock:
     release: "1.el8"
     before: "Package[puppetdb-termini]"

-foreman::repo::repo: "3.8"
-foreman::version: "3.8.0"
+foreman::repo::repo: "3.9"
+foreman::version: "3.9.1"
 puppetdb::globals::version: '7.16.0'

 profile::core::common::manage_sssd: false
```

site/profile/data/common.yaml
```diff
@@ -8,7 +8,7 @@ profile::core::convenience::packages:
   - "neovim"
   - "parallel"
   - "rsync"
-  - "screen"
+#  - "screen" not available on EL8
```
jhoblitt: It is present in epel.  https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/s/  Maybe the tags you used prevented epel from being setup?

site/profile/manifests/core/common.pp
```diff
@@ -67,7 +71,6 @@
   include profile::core::ca
   include profile::core::convenience
   include profile::core::dielibwrapdie
-  include profile::core::ipa
```
jhoblitt: We should probably add a param for all of the auth (ipa, krb5, sssd).

site/profile/data/common.yaml
bastelfreak: Oh indeed. It was probably because of the tags.

bastelfreak: you're right. the screen installation still works.

PR #1055, opened 2024-02-20
(it-ztp.cp) EL9 Redeployment

dtapiacl:
This PR was created to convert the ZTP deployment to EL9 along with its network configuration.

### DIFF
--- a/hieradata/node/it-ztp.cp.lsst.org.yaml
+++ b/hieradata/node/it-ztp.cp.lsst.org.yaml
@@ -1,0 +1,18 @@
+---
+nm::connections:
+  ens192:
+    content:
+      connection:
+        id: "ens192"
+        uuid: "03da7500-2101-c722-2438-d0d006c28c73"
+        type: "ethernet"
+        interface-name: "ens192"
+      ethernet: {}
+      ipv4:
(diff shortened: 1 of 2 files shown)


jhoblitt (approved):
LGTM.

hieradata/node/it-ztp.cp.lsst.org.yaml
```diff
@@ -0,0 +1,18 @@
...
+        type: "ethernet"
+        interface-name: "ens192"
+      ethernet: {}
+      ipv4:
+        address1: "10.17.9.253/24,10.17.9.254"
+        dns: "139.229.160.53;139.229.160.54;"
+        dns-search: "cp.lsst.org;"
+        method: "manual"
+      ipv6:
+        method: "ignore"
```
jhoblitt: I think we should probably be using `disabled` here.  I don't recall why we ended up with a mix of `disabled` and `ignore.  I think older versions of NM might not have supported `disabled`.

jhoblitt (changes requested):
Lets start using `disabled` consistently unless there is a reason to do otherwise.

jhoblitt (approved):
LGTM.

PR #1059, opened 2024-02-20
(ccs) do not prefix sal services with instrument- on lsstcam-mcm hosts

glennmorris:
For historical reasons, the ccs SAL services use an instrument- prefix (eg comcam-) on comcam and auxtel, but on lsstcam no prefix is used.


jhoblitt:
@glennmorris Rebasing on current production should fix the CI failure.

jhoblitt (approved):
LGTM.

PR #1060, opened 2024-02-20
(gha) use 32core runner for "rake checks"

jhoblitt:
Reduces run time from ~8 mins -> ~5mins.


jhoblitt:
I'm unsure of how much of win this ends up being as it appears to take > 1min for a 32c runner to come online.

jhoblitt:
This reduces the current time from ~6mins to ~4mins. However, the total wallclock time seems closer to 6mins as it seems to take longer to get a 32c runner.

PR #1061, opened 2024-02-20
(role/ccs-mcm) move common services here from mcm nodes

glennmorris:
This moves mmm, cluster-monitor, lockmanager to the role. On lsstcam-mcm.ls, it adds mmm (previously missing). On lsstcam-mcm.cp, it adds all three (previously missing).

Affects nodes:
comcam-mcm.{cp,tu}
lsstcam-mcm.{cp,ls}


jhoblitt (approved):
:+1:  I'm very happy to see progress on moving the service configuration into roles.

PR #1062, opened 2024-02-21
(auxtel-dc01.tu.lsst.org) EL9 network conversion

dtapiacl:
Add EL9 network conversion for auxtel-dc01.tu.lsst.org ex forwarders.tu.lsst.org.


hieradata/node/auxtel-dc01.tu.lsst.org.yaml
```diff
@@ -0,0 +1,34 @@
+---
+profile::daq::daq_interface::hwaddr: "a0:36:9f:20:4a:76"
+profile::daq::daq_interface::uuid: "dced8e46-f782-472a-b6a8-b7876a22afff"
+profile::daq::daq_interface::was: "p2p2"
+profile::daq::daq_interface::mode: "dhcp-client"
```
jhoblitt: All of the `profile::daq::*` keys can be removed.  That was EL7 specific configuration.

hieradata/node/auxtel-dc01.tu.lsst.org.yaml
```diff
@@ -0,0 +1,34 @@
...
+    content:
+      connection:
+        id: "ens2f1"
+        uuid: "da97db8e-6400-4cca-9334-71995eb2ac1d"
+        type: "ethernet"
+        interface-name: "ens2f1"
+      ethernet: {}
+      ipv4:
+        method: "manual"
+        address1: "192.168.100.13/24"
```
jhoblitt, line 41: DAQ clients have always been on dhcp.  Only the daq-mgt node, which acts as the DAQ network dhcp server, should have a static IP.

In the BTS, we have been creating a bridge interface named `lsst-daq` and I think we should probably try to standardize onto that configuration.

 See:

- https://github.com/lsst-it/lsst-control/blob/production/hieradata/site/ls/variant/1114s/subvariant/daq-lhn.yaml#L3-L19
- https://github.com/lsst-it/lsst-control/blob/production/hieradata/site/ls/variant/1114s/subvariant/daq-lhn.yaml#L80-L99

jhoblitt (changes requested):

hieradata/node/auxtel-dc01.tu.lsst.org.yaml
dtapiacl: I see, to clear things up. Since this node is a DAQ client its interface should be set to DHCP. For the DAQ-MGT I should create a bridge interface named lsst-daq. I will remove the static configuration from this node and set it to DHCP.

hieradata/node/auxtel-dc01.tu.lsst.org.yaml
dtapiacl: I will cleaned this up.

hieradata/node/auxtel-dc01.tu.lsst.org.yaml
```diff
@@ -0,0 +1,28 @@
...
+        uuid: "03da7500-2101-c722-2438-d0d006c28c73"
+        type: "ethernet"
+        interface-name: "ens2f0"
+      ethernet: {}
+      ipv4:
+        method: "auto"
+      ipv6:
+        method: "disabled"
+      proxy: {}
+  ens2f1:
```
jhoblitt, line 16: There needs to be an interface named `lsst-daq` as this is used in the ccs/daq configuration.

jhoblitt (changes requested):

hieradata/node/auxtel-dc01.tu.lsst.org.yaml
jhoblitt: All hosts connected to the daq require an interface named `lsst-daq`.

jhoblitt (approved):
LGTM.

PR #1066, opened 2024-02-22
(ztp.ls.lsst.org) EL9 Redeployment.

dtapiacl:
IHS-4998

### DIFF
--- a/hieradata/node/ztp.ls.lsst.org.yaml
+++ b/hieradata/node/ztp.ls.lsst.org.yaml
@@ -1,0 +1,18 @@
+---
+nm::connections:
+  ens192:
+    content:
+      connection:
+        id: "ens192"
+        uuid: "03da7500-2101-c722-2438-d0d006c28c73"
+        type: "ethernet"
+        interface-name: "ens192"
+      ethernet: {}
+      ipv4:
+        address1: "10.49.9.253/24,10.49.9.254"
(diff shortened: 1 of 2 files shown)


hieradata/node/it-ztp.ls.lsst.org.yaml
```diff
@@ -0,0 +1,18 @@
...
+    content:
+      connection:
+        id: "ens192"
+        uuid: "03da7500-2101-c722-2438-d0d006c28c73"
+        type: "ethernet"
+        interface-name: "ens192"
+      ethernet: {}
+      ipv4:
+        address1: "10.49.9.253/24,10.49.9.254"
+        dns: "139.229.135.53;139.229.135.54;"
```
jhoblitt: We should set all 3 forward resolvers.

jhoblitt (changes requested):
Can we rename this host to just `ztp`?

hieradata/node/it-ztp.ls.lsst.org.yaml
dtapiacl: Included missing resolver.

dtapiacl:
> Can we rename this host to just `ztp`?

I don't see why not, host was renamed to ztp.ls.lsst.org.

jhoblitt (approved):
LGTM.  Thank you for changing the hostname.

PR #1073, opened 2024-03-06
(site/ls) bump docker to 24.0.9

jhoblitt:
All ls rke nodes have already been upgraded to 24.0.9.  At least these non-rke hosts have docker as part of their role:

-  azar01.ls.lsst.org 
-  foreman.ls.lsst.org 
-  love01.ls.lsst.org 
-  tel-hw1.ls.lsst.org


cbarria (approved):
Looks Great, Thanks!

PR #1074, opened 2024-03-06
(site/cp) bump docker to 24.0.9

jhoblitt:
Note that some tu hosts are still on docker 23.x and will need to be upgraded before the default docker version may be changed to 24.x.


dtapiacl (approved):
LGTM.

PR #1076, opened 2024-03-08
(node/comcam-mcm.tu.lsst.org) EL9 deployment

dtapiacl:


jhoblitt (changes requested):
The yaml looks good.  The commit / PR title should for this repo should state what resource is being changed. E.g. `(node/comcam-mcm.tu.lsst.org)` because there are so many things would could impact a single host.

jhoblitt (approved):
I have fixed the issues with the tests failing. The spec had the wrong interface name for the `dds` interface and the `cluster`  `node_params` needed to be set.

PR #1079, opened 2024-03-11
(node/auxtel-dc01.tu.lsst.org) enable dhcp on lsst-daq interface

jhoblitt:
All daq clients should use dhcp to configure the `lsst-daq` interface.


glennmorris (commented):
Looks alright to me; but I will leave others to approve, since I remember some concern about dhcp being used for addresses that people would prefer to be static (eg so that systems come up correctly when dhcp is not available).

dtapiacl (approved):
LGTM.

jhoblitt:
@glennmorris DAQ clients have always been on dhcp. Only the DAQ management node has had a static address as it is the dhcp server for that network segment.

PR #1080, opened 2024-03-11
(node/daq-mgt.tu.lsst.org) set lsst-daq int rx/tx ring to 4096

jhoblitt:


hieradata/node/daq-mgt.tu.lsst.org.yaml
dtapiacl, line 1: I left it at 4000 originally, which is what we discussed last time, but I inputted the wrong buffer size increment value.

dtapiacl (approved):

hieradata/node/daq-mgt.tu.lsst.org.yaml
jhoblitt: I'd like to fix the value now so we don't have to think about it in the future.

PR #1082, opened 2024-03-11
(node/comcam-archiver.tu.lsst.org) rm dead el7 network config

jhoblitt:


dtapiacl (commented):
Network configuration needs to be converted to EL9 correct ?

jhoblitt:
It is probably fine to leave the host on dhcp with no configuration.  NM enables dhcp on all interfaces when there is no configuration.  This host is only an NFS server and doesn't need to communicate with dds anymore.

dtapiacl (approved):

PR #1083, opened 2024-03-11
(spec) fix misc spec issues for comcam-*.tu hosts

jhoblitt:


dtapiacl (approved):
Thank you for adding those values. It looks good to me.

PR #1087, opened 2024-03-11
(role/comcam-fp) move ccs services here from the node level

glennmorris:
The net result is to add the service comcam-daq-monitor on comcam-fp01.tu.


jhoblitt (approved):
LGTM.

PR #1088, opened 2024-03-12
(site/dev) send syslog to opensearch.ayekan.dev.lsst.org

jhoblitt:
Note that the concat resource used by `puppet/rsyslog` to template the rsyslog config file will leak secrets into the agent logs.  The `puppet/rsyslog` module needs a param added to allow diffs to be disabled.

Related to https://github.com/lsst-it/lsst-puppet-hiera-private/pull/109

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -39,6 +39,12 @@ lookup_options:
   ipset::sets:
     merge:
       strategy: "deep"
+  rsyslog::feature_packages:
+    merge:
+      strategy: "deep"
+  rsyslog::config::modules:
+    merge:
+      strategy: "deep"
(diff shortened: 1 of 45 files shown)


cbarria (approved):
Looks Great 👍🏻

jhoblitt:
Rebased on master and the `# XXX this is super ugly, we may need to pass in site as a param` old comment  in the rsyslog related spec has been addressed.

fbegyn (approved):
LGTM. My puppet knowledge doesn't reach far enought to see where exactly the auth is handled or merged from, but I see the correct traces.

jhoblitt:
The auth parameters are coming from eyaml in another repo.

PR #1091, opened 2024-03-18
(node/auxtel-mcm.tu.lsst.org) EL9 upgrade

dtapiacl:
Ticket of reference: https://rubinobs.atlassian.net/browse/IT-5100


jhoblitt (approved):
LGTM but why is WIP in the PR title?

PR #1092, opened 2024-03-18
(node/auxtel-archiver.tu.lsst.org)  EL9 upgrade

dtapiacl:
Ticket of reference: https://rubinobs.atlassian.net/browse/IT-5102


jhoblitt (approved):
This is OK as is but note that the `dds` interface is no longer required.

PR #1093, opened 2024-03-20
(profile::ccs::file_transfer) add option to install s3daemon

glennmorris:
New class parameters:
(s3daemon) boolean, true to install
(s3daemon_repo_url, s3daemon_repo_rev, s3daemon_repo_directory)
  strings for repo URL and revision, and install location
(s3daemon_env_file, s3daemon_env_url, s3daemon_env_access, s3daemon_env_secret)
  strings for environment file location, and contents

When true: install required yum and pip packages; clone repository; configure and start s3daemon system service.

(templates/ccs/service.epp) new file
(templates/ccs/file_transfer/s3daemon_envfile.epp) new file

### DIFF
--- a/site/profile/manifests/ccs/file_transfer.pp
+++ b/site/profile/manifests/ccs/file_transfer.pp
@@ -1,8 +1,6 @@
 # @summary
 #   Install items related to ccs file transfer.
 #
-# @param install
-#   Boolean, false means do nothing.
 # @param directory
 #   String specifying directory to install to.
 # @param user
@@ -27,7 +25,6 @@
 #   String specifying password for pkgurl
 #
 class profile::ccs::file_transfer (
-  Boolean $install = false,
   String $directory = '/home/ccs-ipa/bin',
   String $user  = 'ccs-ipa',
   String $group = 'ccs-ipa',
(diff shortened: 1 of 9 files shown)


glennmorris:
This requires:
https://github.com/lsst-it/lsst-puppet-hiera-private/pull/110

site/profile/manifests/ccs/file_transfer.pp
```diff
@@ -39,8 +55,132 @@
   String $pkgurl = $profile::ccs::common::pkgurl,
   String $pkgurl_user = $profile::ccs::common::pkgurl_user,
   String $pkgurl_pass = $profile::ccs::common::pkgurl_pass,
+  Boolean $s3daemon = false,
+  Stdlib::HTTPUrl $s3daemon_repo_url = 'https://github.com/lsst-dm/s3daemon',
+  Optional[String] $s3daemon_repo_rev = undef,
+  String $s3daemon_repo_directory = '/home/ccs-ipa/s3daemon/git',
```
jhoblitt: It is a good idea to use `String[1]` instead of `String` unless there is a need to support `""`.

site/profile/manifests/ccs/file_transfer.pp
```diff
@@ -39,8 +55,132 @@
...
+  String $s3daemon_env_file = '/home/ccs-ipa/s3daemon/env',
+  String $s3daemon_env_url = 'https://s3dfrgw.slac.stanford.edu',
+  String $s3daemon_env_access = 'access-key',
+  String $s3daemon_env_secret = 'secret-key',
```
jhoblitt: I assume these are s3 credentials?  Shouldn't these be mandatory parameters?

site/profile/manifests/ccs/file_transfer.pp
```diff
@@ -39,8 +55,132 @@
...
 ) {
   if $install {
+    if $s3daemon {
```
jhoblitt: The nested `if` here is a bit ugly.  Can this new functionality be in a new profile or all of this moved to a stand alone module?  (also, why does the `install` param exist?)

jhoblitt (changes requested):

site/profile/manifests/ccs/file_transfer.pp
glennmorris: Probably the install parameter exists because I wanted to reuse some parameters from profile::ccs::common, and thought that meant I should include profile::ccs::file_transfer in common, rather than having it be entirely separate. I then needed a way to control which hosts it was applied on. But I suppose it is reasonable to assume that profile::ccs::common will always be in use, so I could indeed make this separate and get rid of install.

In the short/medium term, I believe s3daemon will replace the old way of doing file transfer, so hopefully it is only a temporary situation to need two separate flags.

site/profile/manifests/ccs/file_transfer.pp
glennmorris: I suppose that would be clearer, yes.

glennmorris:
@jhoblitt thanks for the comments. I have pushed a new version - please re-review when you have time, thanks.

jhoblitt (approved):
LGTM.  Thank you for implementing my comments.

PR #1094, opened 2024-03-20
(role/htcondor) installs htcondor scheduler

cbarria:
HTCondor Scheduler profile


jhoblitt:
@cbarria Do you want review on this PR while its a draft and (WIP)?

cbarria:
not yet, i might noodle with github in the wrong way today... we have the module class still pending, sorry.

jhoblitt:
@cbarria FYI - Even if a PR is set as a draft, when you add reviewers the request shows up in https://github.com/pulls/review-requested.

hieradata/site/cp/role/htcondor.yaml
```diff
@@ -0,0 +1,19 @@
+---
+classes:
```
jhoblitt: It appears that `spec/unit/top_level_role_spec.rb` was broken by the migration to eyaml.  It should have failed on top level `htcondor` role not existing.  Our pattern is that there must be a top level role.  The classes/etc. should be factored out to a common top level role.

jhoblitt (changes requested):

hieradata/site/cp/role/htcondor.yaml
jhoblitt: Fix for the broken spec: https://github.com/lsst-it/lsst-control/pull/1118

cbarria: I need more specifics on this as I don't fully understand what it means, I will ask later, thanks!

jhoblitt: `hieradata/role/htcondor.yaml` is required to exist.

jhoblitt: #1118 has been merged.  When this branch is rebased on `production`, there should be a test failure now.

cbarria: yeah it broke but i added the file and now its fine 👍🏻

hieradata/site/cp/role/htcondor.yaml
```diff
@@ -0,0 +1,14 @@
+---
+htcondor::htcondor_version: "23.0"
```
jhoblitt: It looks like the version can be factored out to the top level class too.

cbarria: your right, versions should stay the same and if it needs a different on dev, it can be overwritten at site

jhoblitt (approved):

PR #1097, opened 2024-03-21
(profile::ccs::file_transfer) handle /home/ccs-ipa being absent

glennmorris:
-

### DIFF
--- a/site/profile/manifests/ccs/file_transfer.pp
+++ b/site/profile/manifests/ccs/file_transfer.pp
@@ -61,6 +61,15 @@ class profile::ccs::file_transfer (
   String $pkgurl_user = $profile::ccs::common::pkgurl_user,
   String $pkgurl_pass = $profile::ccs::common::pkgurl_pass,
 ) {
+  $parent = "${dirname($directory)}"
+
+  ensure_resource('file', $parent, {
+      ensure => directory,
+      owner  => $user,
+      group  => $group,
+      mode   => '0755',
+  })
+
   ## We expect that s3daemon will become the default (and only) method
   ## some time soonish.
   if $s3daemon {


site/profile/manifests/ccs/file_transfer.pp
```diff
@@ -180,10 +189,11 @@
...
   file { $directory:
-    ensure => directory,
-    owner  => $user,
-    group  => $group,
-    mode   => '0755',
+    ensure  => directory,
+    owner   => $user,
+    group   => $group,
+    mode    => '0755',
+    require => File[$parent],
```
jhoblitt: This isn't needed as the file type will `autorequire` the parent dir.

site/profile/manifests/ccs/file_transfer.pp
```diff
@@ -61,6 +61,15 @@
   String $pkgurl_user = $profile::ccs::common::pkgurl_user,
   String $pkgurl_pass = $profile::ccs::common::pkgurl_pass,
 ) {
+  $parent = "${dirname($directory)}"
+
+  file { $parent:
```
jhoblitt: It would be a good idea to use [`ensure_resource()`](https://forge.puppet.com/modules/puppetlabs/stdlib/reference#ensure_resource-1) here to prevent a possible conflict with other profiles.

jhoblitt (changes requested):
The PR title/commit message reference /home/ipa it should be /home/ccs-ipa.

site/profile/manifests/ccs/file_transfer.pp
glennmorris: Thanks for the comments; I have pushed a new version.

jhoblitt (approved):

PR #1103, opened 2024-03-26
(ccs_monit) change default mailserver to localhost

glennmorris:
Rather than explicitly listing a remote mailserver, just let localhost handle the mail.


hieradata/site/tu/cluster/auxtel-ccs.yaml
```diff
@@ -12,4 +12,6 @@ clustershell::groupmembers:
   all: {group: "all", member: "auxtel-fp01,auxtel-mcm"}
 
 postfix::relayhost: "[mail.lsst.org]:submission"
-ccs_monit::mailhost: "mail.lsst.org"
+ccs_monit::mailhost:
+  - "mail.lsst.org"
```
jhoblitt: Why is tu still listing an explicit mail relay?

glennmorris: Because it is working with the current configuration, so I left it alone (due to changes in the module, the hiera has to change to preserve the status quo).

jhoblitt: Shouldn't it work using the localhost for relay too?  It seems like this is a configuration difference between a test stand and the summit that could be eliminated.

glennmorris: I'll try it out using localhost and report back.

glennmorris: I have verified that just using localhost in site/tu works fine too.
I have forced pushed a corresponding new version, please review when convenient, thanks.

jhoblitt (approved):
LGTM.

PR #1105, opened 2024-03-27
(node/fiber-spec01.tu.lsst.org) EL9 setup

dtapiacl:


hieradata/node/fiber-spec01.tu.lsst.org.yaml
```diff
@@ -0,0 +1,47 @@
...
+      ethernet: {}
+      bridge:
+        stp: false
+      ipv4:
+        address1: "140.252.147.137/27"
+        method: "manual"
+      ipv6:
+        method: "disabled"
+      proxy: {}
+network::mroutes_hash:
```
jhoblitt: The `example42/network` module isn't included on EL9 and thus this key won't do anything.

jhoblitt (changes requested):
Please remove the noop hiera key and/or add the nm route configuration if it is needed.

hieradata/node/fiber-spec01.tu.lsst.org.yaml
dtapiacl: You are right, it is only supported up to RHEL8. I will just remove it for now as this is just a test device for Parker.

jhoblitt (approved):
Looks good now. Thanks.

PR #1110, opened 2024-03-28
(profile::core::node_info) fwv

jhoblitt:
This empty class exists solely as a vehicle for shipping enc data to puppetdb.

Related to https://github.com/lsst-it/k8s-cookbook/pull/300


cbarria (approved):
Looks Great 👍🏻

PR #1113, opened 2024-04-02
(role/ccs-mcm) install tomcat library jar files

glennmorris:


site/profile/manifests/ccs/tomcat.pp
```diff
@@ -80,6 +85,19 @@
     }
   }
 
+  unless (empty($jars)) {
```
jhoblitt: The unless isn't needed here as `{}.each` is a noop.

jhoblitt (changes requested):

site/profile/manifests/ccs/tomcat.pp
glennmorris: That's what I get for reusing the [existing style](https://github.com/lsst-it/lsst-control/blob/production/site/profile/manifests/ccs/tomcat.pp#L74)...

I have pushed a revised version.

jhoblitt: Ya. You can't trust that guy.

jhoblitt (approved):
LGTM.

tony-johnson (commented):
While we are thinking about this, lets update to the latest version of CCSWebTrending. I will look up what that is.

glennmorris:
> While we are thinking about this, lets update to the latest version of CCSWebTrending. I will look up what that is.

We are already using the latest release in nexus (2.1).

tony-johnson:
> > While we are thinking about this, lets update to the latest version of CCSWebTrending. I will look up what that is.
> 
> We are already using the latest release in nexus (2.1).

Yes, but that is because I have been cheating and pushing out changes without using puppet or the release process. I have just created a 2.2 release.

tony-johnson (approved):
Looks good, thanks

PR #1115, opened 2024-04-03
(role/puppetdb) enable ldap auth

jhoblitt:
The listenings ports are now:

* 80 - redirect to port 443
* 443 -  puppetboard w/ ldap auth
* 8080 - puppetdb w/o auth; loopback only
* 8081 - puppetdb w/ x509 auth
* 8443 - puppetdb w/ ldap auth

Depends on https://github.com/lsst-it/lsst-puppet-hiera-private/pull/114
Related to https://github.com/lsst-it/k8s-cookbook/pull/304

### DIFF
--- a/hieradata/role/puppetdb.yaml
+++ b/hieradata/role/puppetdb.yaml
@@ -3,15 +3,19 @@ classes:
   - "profile::core::common"
   - "profile::core::docker"
   - "profile::core::docker::prune"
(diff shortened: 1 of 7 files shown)


cbarria (approved):
got dizzy by going back to profile::core:puppetdb but then I notice is including the module also inside, looks great 👍🏻

PR #1116, opened 2024-04-08
(site/dev) rm rsyslog -> opensearch

jhoblitt:
As we are already pushing syslog to fluentbit, sending direct to openserach is both redundant and broken.  Resolves this error message:

    2024-04-08T15:57:05.883093+00:00 rancher03.dev.lsst.org rsyslogd: omelasticsearch: error in elasticsearch reply: bulkmode insert does not return array, reply is: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"}],"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"},"status":400} [v8.2102.0-117.el9 try https://www.rsyslog.com/e/2218 ]

### DIFF
--- a/hieradata/site/dev.yaml
+++ b/hieradata/site/dev.yaml
@@ -16,21 +16,6 @@ rsyslog::config::actions:
       StreamDriver: "ossl"
       StreamDriverMode: "1"
       StreamDriverAuthMode: "anon"
-  opensearch:
-    type: "omelasticsearch"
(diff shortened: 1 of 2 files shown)


cbarria (approved):
Looks Great 👍🏻

PR #1118, opened 2024-04-08
(spec) fix spec/unit/top_level_role_spec.rb

jhoblitt:
This spec was broken by the migration to eyaml.


cbarria (approved):
Looks Great 👍🏻

PR #1122, opened 2024-04-09
(site/cp/cluster/rancher) add network config

jhoblitt:
To support multiple networks during the migration to `IT-CORE-SERVICES`.


dtapiacl (approved):
LGTM.

PR #1124, opened 2024-04-10
(node/lsstcam-mcm.ls.lsst.org) update network to pure yaml format

dtapiacl:


jhoblitt (approved):
This looks good.  Note that the format I normal use for commit messages/PR titles is `(<hierarchy>/<layer name>)`. Following that pattern would make the commit message `(node/lsstcam-mcm.ls.lsst.org) update network to pure yaml format`.

PR #1129, opened 2024-04-18
(node/foreman.cp) increase dhcp pool vlan 1310

julioconstanzo:
dhcp pool increased to allow range 139.229.168.140 - 139.229.168.189


dtapiacl (approved):
LGTM.

PR #1132, opened 2024-04-24
(role/foreman) disable Foreman_global_parameter[disk_enc_tang_servers]

jhoblitt:
As this resource isn't idempotent.  However, the configuration is retained as commented out as this configuration is required when re-provisioning a foreman host.


cbarria (approved):
Looks Great, Thanks! 👍🏻

PR #1133, opened 2024-04-25
(profile/archive) remove sysctl:lhn

cbarria:
there were some corrective issues when running Puppet on the archives, so we decided to remove the class from the archive profile

### DIFF
--- a/spec/hosts/roles/auxtel_archiver_spec.rb
+++ b/spec/hosts/roles/auxtel_archiver_spec.rb
@@ -25,7 +25,6 @@ describe "#{role} role" do
           it { is_expected.to compile.with_all_deps }
 
           include_examples 'common', os_facts: os_facts, site: site
-          include_examples 'lhn sysctls'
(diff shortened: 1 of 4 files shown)


jhoblitt (approved):

spec/hosts/roles/auxtel_archiver_spec.rb
```diff
@@ -25,7 +25,6 @@
           it { is_expected.to compile.with_all_deps }
 
           include_examples 'common', os_facts: os_facts, site: site
-          include_examples 'lhn sysctls'
```
jhoblitt, line 28: The shared examples also need to be removed from `comcam_archiver_spec.rb`

cbarria: yes, thanks! I was boarding so I had to close the laptop but I saw the rake fail on my cellphone, fixing it right now

PR #1134, opened 2024-04-25
(profile::core::fiberspec) add spectrometer udev rule

dtapiacl:
Ticket of reference: https://rubinobs.atlassian.net/browse/IT-5290


jhoblitt (approved):
LGTM.

PR #1135, opened 2024-04-26
(core[1-3].cp.lsst.org) clean up old cores

dtapiacl:
- Foreman and IPAM will be cleaned up once PR is approved.
- These servers were brought down to the base on Wednesday.


jhoblitt (approved):
Thanks for cleaning up after me... :)

PR #1136, opened 2024-04-26
(role/htcondor) update config and dry up tests

cbarria:
update puppet role to match requests, also tried to dry up the tests.

### DIFF
--- a/hieradata/site/cp/role/htcondor.yaml
+++ b/hieradata/site/cp/role/htcondor.yaml
@@ -3,11 +3,31 @@ htcondor::htcondor_host: "htcondor-cm.cp.lsst.org"
 
 nfs::client_enabled: true
 nfs::client_mounts:
-  /mnt/project:
+  /project:
     share: "project"
     server: "nfs1.cp.lsst.org"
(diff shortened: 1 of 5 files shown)


dtapiacl (approved):
Lots of mounts on that system if its approved it LGTM.

hieradata/site/cp/role/htcondor.yaml
```diff
@@ -3,11 +3,31 @@ htcondor::htcondor_host: "htcondor-cm.cp.lsst.org"
 
 nfs::client_enabled: true
 nfs::client_mounts:
-  /mnt/project:
+  /project:
```
jhoblitt, line 6: Why is the mount being changed to be `/project`?

cbarria: everyone complained that the environment needs to be the same as other instances they have for the software stack... there is a huge discussion on this on the htcondor channel, it's part of the things I asked people to test on dev but nobody gave feedback to do it while this was been developed.

jhoblitt (approved):

PR #1138, opened 2024-05-02
(role/fiberspec) add dco class to fiberspec role

dtapiacl:


jhoblitt (changes requested):
Please add spec coverage.

dtapiacl:
> Please add spec coverage.

Thanks for the feedback i would appreciate your review once more.

spec/classes/core/fiberspec_spec.rb
```diff
@@ -8,6 +8,7 @@
       let(:facts) { os_facts }
 
       it { is_expected.to compile.with_all_deps }
+      it { is_expected.to contain_class('profile::ts::dco') }
```
jhoblitt: It looks like there are more examples of testing the catalog for a `profile::...` class than I thought there were.  Generally, within a puppet module the goal is test the the resources that end up in catalog and not being sensitive to the internal construction of the module (classes, defines, etc.). The same rule applies for the profiles within our site module.  There are exceptions to the general rule but this case probably shouldn't be an exception.  

Note that as the `fiberspec` example group is being include 2 lines down so this spec test would also be a duplicate.

spec/support/spec/fiberspec.rb
```diff
@@ -2,6 +2,7 @@
 
 shared_examples 'fiberspec' do
   it do
+    is_expected.to contain_class('profile::ts::dco')
```
jhoblitt: The `support/spec/*` shared examples have be used to test the functionality essentially within a single profile.  They certainly could be used for a role but we haven't been doing that as I think it wouldn't compose very well.  I don't think this shared example should test for "dco" resources in the catalog.

site/profile/manifests/core/fiberspec.pp
```diff
@@ -2,6 +2,7 @@
 #   Manages Avantes Fiberspectrometer USB devices.
 #
 class profile::core::fiberspec {
+  include profile::ts::dco
```
jhoblitt: This is duplicating the class inclusion in the role.  While we do sometimes include profile classes from profile classes there isn't a compelling reason to do so here.

hieradata/role/fiberspec.yaml
```diff
@@ -2,3 +2,4 @@
 classes:
```
jhoblitt, line 2: It appears that I messed when reviewing https://github.com/lsst-it/lsst-control/pull/1134 as there is not a spec file covering this role (there is an auto-magic catchall role test for that cases but it isn't really enough coverage).

Please add a `spec/hosts/roles/fiberspec_spec.yaml`.  Within that file, please include the `dco` shared example group. E.g. `include_examples 'dco'`

jhoblitt (changes requested):
It appears that I messed when reviewing https://github.com/lsst-it/lsst-control/pull/1134 as there is not a spec file covering this role (there is an auto-magic catchall role test for that cases but it isn't really enough coverage).

Please add a `spec/hosts/roles/fiberspec_spec.yaml`.  Within that file, please include the `dco` shared example group. E.g. `include_examples 'dco'`

spec/hosts/roles/fiberspec_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      lsst_sites.each do |site|
+        fqdn = "#{role}.#{site}.lsst.org"
+        override_facts(os_facts, fqdn: fqdn, networking: { fqdn => fqdn })
+
+        describe fqdn, :sitepp do
+          let(:site) { site }
+
+          it { is_expected.to compile.with_all_deps }
+
+          include_examples 'dco'
```
jhoblitt, line 29: I think we also want `include_examples 'fiberspec` here.

jhoblitt (changes requested):

jhoblitt (approved):

PR #1139, opened 2024-05-06
(role/amor.yaml) update network config

dtapiacl:


hieradata/site/cp/role/amor.yaml
```diff
@@ -5,4 +5,4 @@ docker::networks::networks:
     driver: "macvlan"
     subnet: "139.229.170.0/24"
     gateway: "139.229.170.254"
-    options: ["parent=p2p1"]
+    options: ["parent=ens2f0"]
```
jhoblitt, line 8: Please add spec coverage for the change. E.g. https://github.com/lsst-it/lsst-control/blob/production/spec/hosts/nodes/azar01.ls.lsst.org_spec.rb#L73-L85

jhoblitt (changes requested):
Most changes should have spec coverage.

hieradata/site/cp/role/amor.yaml
dtapiacl: Test coverage for the change was incorporated.

spec/hosts/nodes/love1.cp.lsst.org_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'love1.cp.lsst.org', :sitepp do
```
jhoblitt: The hostname is `love01.cp.lsst.org`.

jhoblitt (changes requested):
The hostname should be change to match the real host to avoid confusion in the future. Otherwise, looks good.

jhoblitt (approved):

PR #1143, opened 2024-05-10
(node/perfsonar1.ls.lsst.org) add routes for es.net pS nodes

jhoblitt:


jhoblitt:
As there have been no comments after 5 days, I'm going to go ahead and merge this.

PR #1144, opened 2024-05-13
(node/antu[01-04].ls.lsst.org) cluster configuration

dtapiacl:


jhoblitt (changes requested):
The chassis model needs to be added to https://github.com/lsst-it/lsst-control/blob/production/site/profile/manifests/core/hardware.pp#L18 to configure the bmc.

spec/hosts/nodes/antu01.ls.lsst.org_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      context 'with enp65s0f1' do
+        let(:interface) { 'enp65s0f1' }
+
+        it_behaves_like 'nm enabled interface'
+        it_behaves_like 'nm dhcp interface'
+        it_behaves_like 'nm ethernet interface'
+      end
+
+      context 'with enp65s0f0' do
+        let(:interface) { 'enp65s0f0' }
```
jhoblitt: I'm guessing this is the trunk interface?  If so, it should be in the list of disabled interfaces (we do it that way as we don't want either a static or dhcp address assigned to the interface).

site/profile/manifests/core/hardware.pp
```diff
@@ -42,8 +42,10 @@
     /SSG-640SP-E1CR90/: {
       include ipmi
     }
+    /1115HS-TNR/: {
+      include ipmi
+    }
   }
-
```
jhoblitt, line 46: Why is this whitespace being removed?

dtapiacl: Typo.

spec/hosts/nodes/antu01.ls.lsst.org_spec.rb
dtapiacl: Added the interface, to the disabled list.

jhoblitt (commented):
This code looks good but I don't see IP assignments for these hosts in ipam.  Could you add those?

jhoblitt (approved):
Thank you for updating ipam!

PR #1151, opened 2024-05-14
(cluster/lsstcam-ccs) no longer install old ccs releases

glennmorris:
This will speed up install of new HCUs.


dtapiacl (approved):
LGTM.

PR #1152, opened 2024-05-14
(profile::ccs::tomcat) add ccs webtrending and rest server

glennmorris:
This requires https://github.com/lsst-it/lsst-puppet-hiera-private/pull/123

### DIFF
--- a/site/profile/manifests/ccs/tomcat.pp
+++ b/site/profile/manifests/ccs/tomcat.pp
@@ -7,9 +7,13 @@
 # @param jars
 #   Hash of jar files to install in lib directory: basename, source.
 #
+# @param trending_site
+#   String giving web trending site (ir2, ats, comcam, maincamera)
+#
 class profile::ccs::tomcat (
   Hash[String, Hash] $wars = {},
   Hash[String[1],String[1]] $jars = {},
+  String[1] $trending_site = 'maincamera',
 ) {
   include nginx
 
@@ -75,6 +79,11 @@ class profile::ccs::tomcat (
   #   ,
   #   require => Exec['wait for tomcat'],  # config dir creation
   # }
+
+  tomcat::config::properties::property { 'org.lsst.ccs.web.trending.default.site':
+    catalina_base => $catalina_base,
+    value         => $trending_site,
+  }
 
   unless (empty($wars)) {
     $wars.each |String $n, Hash $conf| {
--- a/site/profile/manifests/ccs/tomcat.pp
+++ b/site/profile/manifests/ccs/tomcat.pp
@@ -10,10 +10,26 @@
 # @param trending_site
 #   String giving web trending site (ir2, ats, comcam, maincamera)
 #
+# @param rest_etc_path
+#   Path to CCS tomcat configuration directory.
+#
+# @param rest_url
+#   String giving URL for the rest server.
+#
+# @param rest_user
+#   Sensitive string giving username for the rest server.
+#
+# @param rest_pass
+#   Sensitive string giving password for the rest server.
+#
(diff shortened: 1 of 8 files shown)


site/profile/manifests/ccs/tomcat.pp
```diff
@@ -76,6 +80,11 @@
   #   require => Exec['wait for tomcat'],  # config dir creation
   # }
 
+  tomcat::config::properties::property { 'org.lsst.ccs.web.trending.default.site':
```
jhoblitt, line 99: We really should have spec coverage for the resources.  It looks like we have no coverage at all of the tomcat resources.

jhoblitt (changes requested):

site/profile/manifests/ccs/tomcat.pp
glennmorris: I know. I didn't create this file, and am not sure I know how to add all the test infrastructure from scratch for it.

jhoblitt: @dtapiacl Can you help?

glennmorris: Probably this is also big enough now that it should be in a separate module?

glennmorris: Have updated this with some more stuff (that at first I had put in the ccs_software module).

glennmorris: @jhoblitt I have added some basic tests; please have a look.
I do not know how to test epp() expansion, or that a concat file contains a given fragment.

jhoblitt: @glennmorris That looks like a good start.  Functions, such as `epp()`, are not present in the compiled catalog, only the result of evaluating the expression is present.  Under rspec-puppet, the thing to test is the parameter of the resource that is using the returned value. E.g.
```ruby
 it { is_expected.to contain_file('/foo').with_content('bar')
```
`concat` & `concat::fragments` are present in the catalog (the literal file is not assembled util the agent applies the catalog).  They can be tested like any other resource. E.g.

- https://github.com/lsst-it/lsst-control/blob/52ec40e55e96d7ae8278abd3adb9c946a35686ed/spec/support/spec/rsyslog.rb#L4
- https://github.com/lsst-it/lsst-control/blob/52ec40e55e96d7ae8278abd3adb9c946a35686ed/spec/spec_helper.rb#L151

glennmorris: I tried in 147d286f, but as I expected it failed, in an opaque way:

> expected that the catalogue would contain File[/etc/ccs/tomcat/statusPersister.properties] with content set to /^hibernate.connection.url=jdbc:mysql:\/\/myurl/ but it is set to #<Sensitive [value redacted]>

glennmorris: Amazingly, a fragment test in 6db4747 seems to have worked, but it relies on knowing the internal implementation details of the puppetlabs-tomcat module.

jhoblitt: Since its an external module, its fair game to look for the define resource directly. E.g.

```ruby
it { is_expected.to contain_tomcat__config__properties__property('org.lsst.ccs.web.trending.default.site')... }
```

but sometimes looking at implementation details is necessary.

glennmorris: Tests are now passing.
(There go several hours I won't get back.)

jhoblitt (approved):
LGTM. Thank you starting on the spec.

PR #1155, opened 2024-05-15
(profile::ccs::tomcat) add ccs webtrending and rest server

glennmorris:
This replaces an earlier version that put the rest server stuff in the ccs_software module.


glennmorris:
Have used this to replace an earlier version of pull #1152 , and am closing this draft.

PR #1156, opened 2024-05-16
(node/lsstcam-vw01.cp.lsst.org) Provision EL9 host

igonzalezcl:


spec/hosts/nodes/lsstcam-vw01.cp.lsst.org_spec.rb
dtapiacl, line 1: Please update the hostname on the test to the correct value.

dtapiacl (dismissed):
Please double check tests for the correct names.

hieradata/node/lsstcam-vw01.cp.lsst.org.yaml
jhoblitt, line 1: It looks like this host doesn't have a device in ipam.  It would be great if we could correct issues like that when reprovisioning a host.

jhoblitt (changes requested):
Please squash 8b14f6ab482dc245278929806ba13a5aa3cdcfd8

jhoblitt (approved):
LGTM. Thank you for updating ipam.

PR #1160, opened 2024-05-20
(rubocop) exclude "vendor"

glennmorris:
Currently only ".vendor" is being excluded, but prior to 81cf968 both were.
.gitignore only ignores "vendor".

Otherwise rubocop descends into the vendor directory, which caused me a lot of problems.


jhoblitt (approved):
ouch.

PR #1161, opened 2024-05-20
(node/lsstcam-shutter01.cp.lsst.org) Deploy host as EL9

dtapiacl:


jhoblitt:
I changed the name of the fqdn interface from eno1 -> enp0s31f6.

jhoblitt (approved):

PR #1168, opened 2024-05-22
(profile::core::yum::lsst_ts_private) allow username to be Sensitive

jhoblitt:
To resolve this error:

    Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Class[Profile::Core::Yum::Lsst_ts_private]: parameter 'username' expects a value of type Undef or String, got Sensitive[String] (file: /etc/puppetlabs/code/environments/cp_production/site/profile/manifests/ts/opensplicedds.pp, line: 10, column: 3) on node sal-dx.cp.lsst.org


cbarria (approved):
Looks Great 👍🏻

PR #1169, opened 2024-05-23
(node/lsstcam-fb01.cp.lsst.org) deploy host as "el7"

dtapiacl:


dtapiacl:
This host behaves the same way as lsstcam-fcs as it uses canbus and will have to be deployed as EL7.

glennmorris:
Added a couple of commits to enable the ccs canbus module.
The Puppetfile bump will lead to a trivial merge conflict at some point, sorry.
It requires a private hiera change too.

glennmorris (approved):

hieradata/node/lsstcam-fb01.cp.lsst.org.yaml
```diff
@@ -0,0 +1,12 @@
+---
+network::interfaces_hash:
+  enp3s0:  # fqdn
+    bootproto: "none"
+    ipaddress: "139.229.175.89"
```
jhoblitt, line 5: This address is missing from ipam.

jhoblitt (changes requested):
This device is missing from ipam.

jhoblitt (approved):

PR #1172, opened 2024-05-23
(profile::ccs::graphical) tweak installed desktop apps

glennmorris:
Add google-chrome, drop zoom.
There was a mild request to add the slack desktop client, but I find slack in a browser works fine on Linux.
The fact that there seems to be no active forge module for the slack client seems like a sign.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -10,7 +10,6 @@ mod 'derdanne/nfs', git: 'https://github.com/lsst-it/puppet-nfs', ref: '6d51f51'
 mod 'duritong/sysctl', git: 'https://github.com/duritong/puppet-sysctl', ref: '847ec1c'  # migrate to herculesteam/augeasproviders_sysctl; https://github.com/duritong/puppet-sysctl/pull/48
 mod 'example42/network', git: 'https://github.com/lsst-it/puppet-network', ref: 'c2b5c59'  # allow stdlib/concat 9.x
 mod 'fervid/snapd', '1.2.1' # 2021-05-31 hreinking: snapd for EAS Raspberry Pi
-mod 'jamesnetherton/google_chrome', '0.10.0'
 mod 'jamtur01/httpauth', '0.0.6'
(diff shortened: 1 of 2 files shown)


glennmorris:
Hmm, puppet librarian not happy with something. It worked when I tried it in practice. :man_shrugging:

glennmorris:
> Hmm, puppet librarian not happy with something. It worked when I tried it in practice. 🤷‍♂️

I guess the root problem is that the new google-chrome module has:
`puppetlabs/apt (>=2.0.0 <8.0.0)`
but Puppetfile has
`mod 'puppetlabs/apt', '9.1.0'`

Do we need to fork the forge chrome module and change the deps?

jhoblitt:
> Do we need to fork the forge chrome module and change the deps?

Yes, unfortunately.  r10k gets us into trouble because it doesn't do recursive dep resolution, which is why there's a librarian-puppet CI check (r10k deploys much faster on the puppetservers though).  It looks like someone else has already started a fork as well: https://github.com/GEANT/puppet-google-chrome

I haven't looked at the changes but that geant fork may be a better starting point for a fork.

jhoblitt:
Maybe split chrome off into its own PR for now so the rest of this can be merged?

glennmorris:
> > Do we need to fork the forge chrome module and change the deps?
> 
> Yes, unfortunately. r10k gets us into trouble because it doesn't do recursive dep resolution, which is why there's a librarian-puppet CI check (r10k deploys much faster on the puppetservers though). It looks like someone else has already started a fork as well: https://github.com/GEANT/puppet-google-chrome
> 
> I haven't looked at the changes but that geant fork may be a better starting point for a fork.

Sadly the geant fork has the same dependency issue.

glennmorris:
I've created a trivial fork of the google-chrome module (not sure of the etiquette) and switched to that.
It passes the tests, and works fine in practice.

jhoblitt (approved):
LGTM.

PR #1181, opened 2024-05-29
(cluster/comcam-ccs) switch to java17

glennmorris:
This removes a tu site-specific setting.


glennmorris:
Converting to draft since we don't want this to get merged before Thursday.

jhoblitt (approved):
I am OK with this is principal but I still have a strong preference for switching to the OS provided java17 packages.

jdk 17.0.3 has multiple known vulnerabilities. https://github.com/lsst-it/puppet-java_artisanal/blob/master/manifests/java17.pp#L23

PR #1196, opened 2024-06-03
(role/rke) rke bump dev to 1.5.9

cbarria:
bump rke versions to move into 1.28.9

### DIFF
--- a/hieradata/site/dev/role/rke.yaml
+++ b/hieradata/site/dev/role/rke.yaml
@@ -1,2 +1,2 @@
 ---
-profile::core::rke::version: "1.5.8"
+profile::core::rke::version: "1.5.9"
--- a/hieradata/role/rke.yaml
+++ b/hieradata/role/rke.yaml
@@ -23,7 +23,6 @@ profile::core::helm::version: "3.5.4"
 profile::core::kernel::debuginfo: true
 profile::core::kernel::devel: true
 profile::core::common::manage_node_exporter: false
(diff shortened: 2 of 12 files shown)


hieradata/site/cp/role/rke.yaml
```diff
@@ -0,0 +1 @@
+profile::core::rke::version: "1.5.8"
```
jhoblitt: The `---` is missing.

jhoblitt (changes requested):
The `(...)` should state that its a role being changed. E.g. `(role/rke)`.  The PR title needs to be updated too.

hieradata/site/cp/role/rke.yaml
cbarria: yeah, my fault, was added.

hieradata/site/dev/role/rke.yaml
```diff
@@ -1,2 +1,2 @@
 ---
-profile::core::rke::version: "1.5.8"
+profile::core::rke::version: "1.5.9"
```
jhoblitt, line 2: Is this actually needed?  It should get the default like the other sites.

hieradata/site/cp/role/rke.yaml
jhoblitt: Is this actually needed? It should get the default like the other sites.

jhoblitt (approved):

PR #1197, opened 2024-06-03
(site/ls) update ccs clusters for cycle 37

glennmorris:
As written, this only affects BTS.
~It installs an EL7 version of OpenSplice-DDS, because at time of writing this is the only flavour of the cycle 37 version (6.10.4-6) that exists. I don't yet know if this will install on EL9 (probably).~
It turns out this was a typo on the cycle 37 page, and no OpenSpliceDDS change is needed.


tony-johnson (approved):
Looks good to me

PR #1199, opened 2024-06-04
(profile::ccs::home) handle EL8+ login.defs home dir perms

glennmorris:
To make new home directories world-readable by default in EL8+, it is necessary to modify HOME_MODE in /etc/login.defs.

### DIFF
--- a/site/profile/manifests/ccs/home.pp
+++ b/site/profile/manifests/ccs/home.pp
@@ -4,7 +4,13 @@ class profile::ccs::home {
   ## rather than each having their own?
   file_line { 'Change default home permissions':
     path  => '/etc/login.defs',
-    match => '^UMASK ',
+    match => '^UMASK\s',
     line  => 'UMASK 022',
+  }
+  file_line { 'Change default home permissions for EL8+':
+    path               => '/etc/login.defs',
+    match              => '^HOME_MODE\s',
+    line               => 'HOME_MODE 0755',
+    append_on_no_match => false,
   }
 }


site/profile/manifests/ccs/home.pp
```diff
@@ -4,7 +4,13 @@
...
   file_line { 'Change default home permissions':
     path  => '/etc/login.defs',
-    match => '^UMASK ',
+    match => '^UMASK\s',
     line  => 'UMASK 022',
   }
+  file_line { 'Change default home permissions for EL8+':
+    path               => '/etc/login.defs',
+    match              => '^HOME_MODE\s',
+    line               => 'HOME_MODE 0755',
```
jhoblitt, line 13: I think this would probably apply to home dirs created for IPA users as well.  What is trying to be accomplished?

glennmorris: Having it apply to IPA users is the wanted behaviour.
The intent is to make the home dirs of normal users world-readable by default, since this is a collaborative enterprise and people need to share work.

jhoblitt: This needs to wait until Cristian returns from leave next week for further discussion.

jhoblitt (changes requested):
This is on hold pending further discussion.

glennmorris:
> This is on hold pending further discussion.

OK.

If you want to keep them private (by default) for some security purposes, we'll live with it, but otherwise historically the CCS team's position was that we are all in this together, and these systems are designed for shared work towards common goals, not private silos.

glennmorris:
PS. the existing code that makes them public on CentOS 7 hosts has been in place for 4 years.
The changes in this pull request are because the syntax for doing that changed in RHEL8.

glennmorris:
So, any thoughts/discussion about this?

glennmorris:
No "further discussion" of any kind has happened in six months. :man_shrugging:

PR #1202, opened 2024-06-04
(node/lsstcam-dc02.cp) disable ccs image-handling service

glennmorris:
This node is a special case of the ccs-dc role.

### DIFF
--- a/hieradata/node/lsstcam-dc02.cp.lsst.org.yaml
+++ b/hieradata/node/lsstcam-dc02.cp.lsst.org.yaml
@@ -20,3 +20,13 @@ nm::connections:
       method=ignore
 
       [proxy]
+
+## The role ccs-dc adds the image-handling service to all lsstcam-dc*.cp nodes.
+## However, dc02.cp is a special case, where the service should not run.
+profile::core::systemd::dropin_file:
+  override.conf:
+    unit: "image-handling.service"
+    content: |
+      [Service]
+      ExecStart=
+      ExecStart=/bin/false


hieradata/node/lsstcam-dc02.cp.lsst.org.yaml
```diff
@@ -20,3 +20,13 @@ nm::connections:
...
+
+## The role ccs-dc adds the image-handling service to all lsstcam-dc*.cp nodes.
+## However, dc02.cp is a special case, where the service should not run.
+profile::core::systemd::dropin_file:
+  override.conf:
+    unit: "image-handling.service"
+    content: |
+      [Service]
+      ExecStart=
+      ExecStart=/bin/false
```
jhoblitt, line 32: What are the semantics of a duplicate key?

glennmorris: > What are the semantics of a duplicate key?

Are you asking about the duplicate ExecStart? If so, before changing this, it is necessary to first reset it to empty, else systemd will complain
`foo.service has more than one ExecStart= setting, which is only allowed for Type=oneshot services. Refusing.
`

jhoblitt: I can't say I've ever tried to that with a service unit.

I think it would be better to not install the service rather than managing the service unit after the fact.  To do that, a knockout prefix would need to be added to https://github.com/lsst-it/lsst-control/blob/production/hieradata/common.yaml#L30 and then on the host something like https://github.com/lsst-it/lsst-control/blob/dfe260b333941f33ac13f440690992b6e7b71906/hieradata/role/foreman.yaml#L185-L186

glennmorris: The override method definitely works, but yes, not installing the service in the first place would be better.
I didn't know how to do that, but will take a look.

glennmorris:
Replacing this with pull #1208

PR #1211, opened 2024-06-10
(role/rke) bump LS client to v1.5.9

cbarria:
bumping clients to update k8s to 1.28.9 =)

### DIFF
--- a/spec/hosts/roles/rke_spec.rb
+++ b/spec/hosts/roles/rke_spec.rb
@@ -35,7 +35,8 @@ shared_examples 'generic rke' do |os_facts:, site:|
     )
   end
 
-  if (site == 'dev') || (site == 'tu')
+  case site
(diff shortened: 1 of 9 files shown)


spec/hosts/roles/rke_spec.rb
```diff
@@ -35,7 +35,7 @@
     )
   end
 
-  if (site == 'dev') || (site == 'tu')
+  if (site == 'dev') || (site == 'tu') || (site == 'ls')
```
jhoblitt: This should probably be a case statement but I realize I'm the one that started if off with an if/or.

jhoblitt (approved):

spec/hosts/roles/rke_spec.rb
cbarria: great suggestion, changed.

PR #1212, opened 2024-06-10
(Gemfile) pin facterdb < 2

jhoblitt:
Facterdb 3.x dropps support for EL7.  facterdb 2.1.0 also seems to fail. Thus, we are pinning facterdb to 1.x until support for EL7 is removed from this repo.


dtapiacl (approved):
LGTM

cbarria (approved):
wanted to fix it so i can learn xD but thanks!

PR #1213, opened 2024-06-10
(hiera/common.yaml) add rsyslog logrotate

cbarria:
are you constantly out of space on your disks? is your node always tainted with disk pressure? is your CNI complaining and your canal pods are unable to start? Well... logs are no longer going to be *THAT* problem, only if you click Approve 👌🏻

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -337,6 +337,29 @@ rsyslog::config::inputs:
       file: "/var/log/audit/audit.log"
       Tag: "auditd"
 
+files:
+  /etc/logrotate.d/rsyslog:
+    ensure: 'file'
+    owner: 'root'
+    group: 'root'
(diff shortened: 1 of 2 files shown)


jhoblitt (commented):
Was this tested on EL7 and EL8?  It will probably work but I don't know how the existing `/etc/logrotate.d.rsyslog` are different.

cbarria:
> Was this tested on EL7 and EL8? It will probably work but I don't know how the existing `/etc/logrotate.d.rsyslog` are different.

this was tested on the foreman dev host with EL7 and some other EL8 cant recall the host 👍🏻

jhoblitt (commented):
This is going to apply to EL7, EL8, EL9, and the couple of Ubuntu nodes.  Has this file been inspected on all of the platforms?

PR #1215, opened 2024-06-11
(node/comcam-vs01.cp.lsst.org) provision host as EL9 & add test

badenerb:


jhoblitt (changes requested):
The code looks good.  The commit message looks a little strange in that the commit title is duplicated as the message of the commit, which was then duplicated into the PR message.  It should be fine to remove the message body to leave only the PR & commit title.

image

jhoblitt (approved):
LGTM.  Thanks for fixing the commit message.

PR #1219, opened 2024-06-11
(node/foreman.cp) elqui dhcp pools for vlan1801 and vlan1802

julioconstanzo:
Add DHCP pool configs for VLAN1801 and VLAN1802

### DIFF
--- a/hieradata/node/foreman.cp.lsst.org.yaml
+++ b/hieradata/node/foreman.cp.lsst.org.yaml
@@ -186,6 +186,20 @@ dhcp::pools:
       - {network: "134.79.23", mask: "24", gateway: "139.229.180.254"}
       - {network: "134.79.235.224", mask: "28", gateway: "139.229.180.254"}
       - {network: "134.79.235.240", mask: "28", gateway: "139.229.180.254"}
+  ELQUI:
+    network: "139.229.181.0"
+    mask: "255.255.255.224"
+    gateway: "139.229.181.30"
+    range:
+      - "139.229.181.21 139.229.181.25"
+    search_domains: "%{alias('dhcp::dnsdomain')}"
+  ELQUI-SVC:
+    network: "139.229.181.32"
+    mask: "255.255.255.224"
+    gateway: "139.229.181.62"
+    range:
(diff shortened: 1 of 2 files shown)


hieradata/node/foreman.cp.lsst.org.yaml
```diff
@@ -186,6 +186,20 @@ dhcp::pools:
       - {network: "134.79.23", mask: "24", gateway: "139.229.180.254"}
       - {network: "134.79.235.224", mask: "28", gateway: "139.229.180.254"}
       - {network: "134.79.235.240", mask: "28", gateway: "139.229.180.254"}
+  ELQUI:
+    network: "139.229.181.0"
+    mask: "255.255.255.224"
+    gateway: "139.229.181.30"
+    range:
+      - "139.229.181.1 139.229.181.20"
```
jhoblitt: The "range" is the set of IPs used for dynamic leases.  The "dhcp reservations" we create from foreman can be within that range but it can cause problems as the dhcp server will happily hand out the reservation address to random dhcp requests.  Thus, we want the range to be small and not overlap with the address space we will use for reservations.  The way isc dhcpd handles this is confusing.

The range should be shrunk down to be only a few IPs.

hieradata/node/foreman.cp.lsst.org.yaml
```diff
@@ -186,6 +186,20 @@ dhcp::pools:
...
+    search_domains: "%{alias('dhcp::dnsdomain')}"
+  ELQUI-SVC:
+    network: "139.229.181.32"
+    mask: "255.255.255.224"
+    gateway: "139.229.181.62"
+    range:
+      - "139.229.181.33 139.229.181.52"
```
jhoblitt: Similar to ELQUI, the range should be shrunk down to be only a few IPs.

jhoblitt (changes requested):

julioconstanzo:
dynamic dhcp ranges were modified

dtapiacl (approved):
LGTM.

jhoblitt (approved):
LGTM.

PR #1220, opened 2024-06-12
(node/lsstcam-shutter[01,02].cp) install ccs_hcu pth450 support

glennmorris:
This adds a udev rule and some binaries in /usr/local/bin


dtapiacl (approved):
LGTM.

PR #1221, opened 2024-06-12
(ccs) cycle 37 updates

glennmorris:
These apply the CCS updates for cycle 37 to all hosts; ie adds the BTS.


jhoblitt (approved):
This looks reasonable.

PR #1222, opened 2024-06-12
(node/comcam-daq-mgt.cp) rebuild host as EL8

dtapiacl:


cbarria (approved):
Looks Great 👍🏻 but next time, don't do a cycle for the disable interface if only has 1 item on the loop.

PR #1223, opened 2024-06-13
(role/rke) bump CP client to v1.5.9

cbarria:
bump the rke client to move k8s version to v1.28.9


dtapiacl (approved):
Looks good to me. 👍🏻

PR #1224, opened 2024-06-13
(node/comcam-mcm.cp) add dds routes to node

dtapiacl:


cbarria (approved):
Looks Great 👍🏻 we need to figure out the difference between doing routes here versus doing routes on tables

PR #1227, opened 2024-06-14
(profile::ccs::tomcat) miscellaneous small fixes

glennmorris:
Ref https://rubinobs.atlassian.net/browse/IT-5446


jhoblitt (approved):
This seems reasonable.

PR #1228, opened 2024-06-17
 (role/{atsccs,ccs-dc,comcam-fp}) apply basic lhn transfer tuning

jhoblitt:


glennmorris:
So IIUC, the net effect will be to remove systcl.d/99-lsst-daq.conf, which contains:
```
net.core.rmem_max=18874368
net.core.wmem_max=18874368
```
and replace these numbers with larger values, namely `536870912`, in a different file?
Which is why profile::ccs::common::sysctls is set to false - please could an explanatory comment be added to the hiera?

This is probably fine, but since investigation to figure out camera networking issues is currently ongoing, it might be good to hold off for a bit.

cbarria:
KT was requesting to add the following values to the sysctl
```
net.core.rmem_max = 536870912
net.core.wmem_max = 536870912
net.ipv4.tcp_rmem = 4096 87380 536870912
net.ipv4.tcp_slow_start_after_idle = 0
net.ipv4.tcp_wmem = 4096 87380 536870912
```
maybe we can put these changes here... he also said something about setting the MTU to a jumbo value =S

cbarria (approved):
Let's Merge this and wait till the settings are tested before doing something else, Looks Great 👍🏻

glennmorris:
> So IIUC, the net effect will be to remove systcl.d/99-lsst-daq.conf, which contains:
> 
> ```
> net.core.rmem_max=18874368
> net.core.wmem_max=18874368
> ```
> 
> and replace these numbers with larger values, namely `536870912`, in a different file? Which is why profile::ccs::common::sysctls is set to false - please could an explanatory comment be added to the hiera?
> 
> This is probably fine, but since investigation to figure out camera networking issues is currently ongoing, it might be good to hold off for a bit.

From the CCS side, we still prefer that networking changes not be applied to summit ccs-dc hosts until the networking issues with the camera are resolved.

glennmorris:
> From the CCS side, we still prefer that networking changes not be applied to summit ccs-dc hosts until the networking issues with the camera are resolved.

Tony says it is fine to install this now.

glennmorris (approved):

PR #1229, opened 2024-06-18
(cluster/lsstcam) enable s3daemon

glennmorris:
Also on comcam-fp01.tu.


glennmorris:
Seems editing the pull title after the fact doesn't propagate to the commit message checker, so will have to close and open a new request.

PR #1233, opened 2024-06-19
(node/lsstcam-pcs02.cp.lsst.org) provision host as EL9 & add test

badenerb:


jhoblitt (commented):
Is the .175.105 address reservation in ipam correct?

image

jhoblitt (approved):

PR #1236, opened 2024-06-24
(node/comcam-fp01.cp) add hosts entry for SLAC s3

cbarria:
this is the fix KT came up to reach the s3 at SLAC, I just puppetized it


jhoblitt (approved):
LGTM.

PR #1237, opened 2024-06-24
(node/hexrot.cp.lsst.org) clean up network configuration

dtapiacl:


jhoblitt (approved):
Looks like there was no spec.  Thanks for cleaning up.

PR #1248, opened 2024-07-01
(Puppet) Remove sysstat package from hosts.

dtapiacl:
To address the high CPU usage on some nodes making hosts unresponsive, it was decided to remove the sysstat package from our hosts.

Ticket of reference: https://rubinobs.atlassian.net/browse/IT-5490

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -98,7 +98,6 @@ mod 'puppet/tang', '0.1.1'
 mod 'puppet/telegraf', '5.2.0'
 mod 'puppet/tuned', '1.0.0'
 mod 'puppet/yum', '7.1.0'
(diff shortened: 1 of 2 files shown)


jhoblitt (commented):
The plan is to cleanup the crontab of > 300 hosts manually?

dtapiacl:
> The plan is to cleanup the crontab of > 300 hosts manually?

Probably not of course, i did not find on our gitrepo the cronjob that manages the sysstat cron. If you could point as to what the best approach would be I would appreciate it.

jhoblitt (dismissed):
Per discussion, cleanup the sysstat package needs to be implemented.

csilva-cl (approved):
looks good

PR #1249, opened 2024-07-01
(role/rke) bump DEV client to v1.5.10

cbarria:
bump rke client to 1.5.10 on Dev, except for ruka, since now it is running rke2.


dtapiacl (approved):
LGTM.

PR #1252, opened 2024-07-03
(profile::core::k8snode) split out of profile::core::rke

jhoblitt:
This is foundational for the creation of an `rke2` role.


dtapiacl (approved):
LGTM.

PR #1255, opened 2024-07-05
(common) add 139.229.181.0/27|vlan1801 to cpfde ipset

jhoblitt:
This is needed to allow elqui nodes to contact tang.


dtapiacl (approved):
LGTM.

PR #1256, opened 2024-07-05
(spec) mv foreman & 'foreman generic' example groups to support file

jhoblitt:
This is a simple cleanup.


cbarria (approved):
Looks Great 👍🏻

PR #1258, opened 2024-07-09
(role/rke) bump CP client to v1.5.10

cbarria:
I moved the default version inside RKE profile to 1.5.10 and removed the `rke::version` references on all sites.
all tests were moved to this version.


dtapiacl (approved):
LGTM.

PR #1264, opened 2024-07-12
(node/rucio.ls.lsst.org) deploy rucio.ls.lsst.org 

dtapiacl:
Requirements per ticket information on IT-5509:

- Install the following rpms, xrootd, xrootd-selinux, xrootd-libs, xrootd-client, xrootd-client-libs, xrootd-server-libs, xrootd-server

- modify the /lib/systemd/system/xrootd@.service and /lib/systemd/system/cmsd@.service and change the User and Group inside these two unit files to saluser and it group (current it is user xrootd)

- Change the owner of directory /etc/xrootd, /var/{log, run, spool}/xrootd to saluser (this probably need to be done every time there is an update of the xrootd rpms)

- Allow user bwhite and wyang to do a) sudo su - saluser and b) sudo systemctl restart xrootd@summit. We will add Sajid Ali to this list of users once he has an account

- Place VM in BTS-LHN network. 

- Mount the filesystems KT mentioned.


hieradata/role/rucio.yaml
```diff
@@ -1,13 +1,6 @@
 ---
 classes:
   - "profile::core::common"
-  - "profile::core::debugutils"
-  - "profile::core::dmz"
+  - "profile::core::nfsclient"
+  - "profile::core::nfsserver"
```
cbarria: why using the nfsserver profile without setting an nfs server? i think you only need the nfsclient

site/profile/manifests/core/rucio.pp
```diff
@@ -2,74 +2,97 @@
...
+    ensure          => 'present',
+    name            => 'XRootD Development Branch RPMs for Enterprise Linux $releasever',
+    baseurl         => 'https://xrootd.web.cern.ch/repo/testing/el$releasever/$basearch',
+    gpgkey          => 'https://xrootd.web.cern.ch/repo/RPM-GPG-KEY.txt',
+    metadata_expire => '1d',
+    repo_gpgcheck   => '1',
+    priority        => '30',
+    gpgcheck        => '1',
+    enabled         => '0',
+  }
```
csilva-cl: this repo is not necessary if its disable, is there a reason to leave it ?

spec/classes/core/rucio_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          ensure: 'present',
+          name: 'XRootD Development Branch RPMs for Enterprise Linux $releasever',
+          baseurl: 'https://xrootd.web.cern.ch/repo/testing/el$releasever/$basearch',
+          gpgkey: 'https://xrootd.web.cern.ch/repo/RPM-GPG-KEY.txt',
+          metadata_expire: '1d',
+          repo_gpgcheck: '1',
+          priority: '30',
+          gpgcheck: '1',
+          enabled: '0',
+        )
```
csilva-cl: it may not be needed

site/profile/manifests/core/rucio.pp
```diff
@@ -1,75 +1,70 @@
...
+    mode   => '0644',
+    owner  => 'saluser',
+    group  => 'saluser',
   }
 
-  #  Install Pip3 Packages
-  package { $pip_packages:
-    ensure   => 'present',
-    provider => 'pip3',
+  file { '/etc/xrootd':
```
cbarria: all the similar resources can be grouped into an array to DRY(don't repeat yourself) up the code...
Ex:
```file { ['/var/log', '/var/run', '/var/spool', '/etc/xrootd']```

**Review by cbarria — APPROVED** [review]
Looks Great 👍🏻

File: hieradata/node/rucio.ls.lsst.org.yaml [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+  ens192:
+    content:
+      connection:
+        id: "ens192"
+        uuid: "810adb94-e2c0-4d41-b66f-1175a5c358ba"
+        type: "ethernet"
+        interface-name: "ens192"
+      ethernet: {}
+      ipv4:
+        address1: "139.229.153.160/24,139.229.153.254"
```
jhoblitt: Why does this host need a static IP?

site/profile/manifests/core/rucio.pp
```diff
@@ -1,75 +1,70 @@
...
+  }
 
-  #  Define PIP Packages
-  $pip_packages = [
-    'wheel',
-    'cryptography',
-    'rucio',
-  ]
+  # Fetch the xrootd.repo file
+  exec { 'fetch_xrootd_repo':
```
jhoblitt: A good rule of thumb is that if `exec` is being used, it probably isn't the correct approach.  The `yumrepo` type should be configure a yum repo.

site/profile/manifests/core/rucio.pp
```diff
@@ -1,75 +1,70 @@
...
+    ]:
+      ensure  => installed,
+      require => Exec['fetch_xrootd_repo'],
+  }
 
-  #  Generate and sign certificate
-  letsencrypt::certonly { $fqdn:
-    plugin      => 'dns-route53',
-    manage_cron => true,
+  file { '/lib/systemd/system/xrootd@.service':
```
jhoblitt: Where does the content of these files come from? An rpm?

site/profile/manifests/core/rucio.pp
```diff
@@ -1,75 +1,70 @@
...
+    ensure => file,
+    mode   => '0644',
+    owner  => 'saluser',
```
jhoblitt, line 20: Why is a service unit owned by `saluser`?

site/profile/manifests/core/rucio.pp
jhoblitt: These repetitive directory resources should be DRY'd up.

site/profile/manifests/core/rucio.pp
```diff
@@ -1,75 +1,70 @@
...
-    'openssl-devel',
-    'systemd-devel',
-    'zlib-devel',
-    'devtoolset-7',
-    'xrootd',
-    'voms',
-  ]
+class profile::core::rucio {
+  # Ensure the GPG key is imported
+  exec { 'import_xrootd_gpg_key':
```
jhoblitt: The `yumrepo` resource is able to import the gpg key.

jhoblitt (changes requested):

site/profile/manifests/core/rucio.pp
dtapiacl: Code was DRY'd  up for better reading.

hieradata/node/rucio.ls.lsst.org.yaml
dtapiacl: Was no logic behind this, i assumed this host would require a private one. Configuration was updated for DHCP.

PR #1266, opened 2024-07-17
(site/tu/cluster/comcam-ccs) fix JAVA_HOME setting for DDS

glennmorris:


tony-johnson (approved):
This looks good for comcam on TTS.
If this works there we should probably deploy it elsewhere, since there does not seem to be  a good reason to keep different setups on different java 17 clusters.

PR #1267, opened 2024-07-19
(role/daq-mgt) Update DAQ to R5-V10.1

jgthayer:
I'd like to stage DAQ R5-V10.1 to all the DAQ management servers in preparation for deploying it. 

I also removed a couple of places where the version had been commented out, and following the pattern for another spec file, removed the explicit version check.


glennmorris (approved):
Looks fine to me.

PR #1296, opened 2024-08-02
(spec) drop EL7 from default list of tested platforms

jhoblitt:
This reduces the time to run the rake checks workflow by ~1/3.


cbarria (approved):
Looks Great 👍🏻

PR #1364, opened 2024-08-05
(*) Migrate spec to ruby3/puppet8

jhoblitt:


Gemfile
```diff
@@ -3,21 +3,20 @@
...
-  gem 'puppet-lint-resource_reference_syntax', '~> 1.2.0',     require: false
-  gem 'puppet-lint-strict_indent-check', '~> 2.1.0',           require: false
-  gem 'puppet-lint-unquoted_string-check', '~> 2.2.0',         require: false
-  gem 'puppet-lint-variable_contains_upcase', '~> 1.4.0',      require: false
-
-  gem 'facterdb', '< 2', require: false # 3.0.0 drops support for EL7
-
-  gem 'r10k',    require: false
-  gem 'toml-rb', require: false # puppet/telegraf
+  gem 'puppet-lint-resource_reference_syntax', '~> 2.0.0',     require: false
```
bastelfreak: that's a dependency of voxpupuli-puppet-lint-plugins and that's a dependency of voxpupuli-test. So you can drop it I think?

Gemfile
```diff
@@ -3,21 +3,20 @@
...
+  gem 'puppet-lint-strict_indent-check', '~> 3.0.0',           require: false
```
bastelfreak: same here

Gemfile
```diff
@@ -3,21 +3,20 @@
...
+  gem 'puppet-lint-unquoted_string-check', '~> 3.0.0',         require: false
```
bastelfreak: same here

Gemfile
```diff
@@ -3,21 +3,20 @@
...
+  gem 'puppet-lint-variable_contains_upcase', '~> 2.0.0',      require: false
```
bastelfreak: same here

site/profile/manifests/ccs/nvidia.pp
```diff
@@ -40,7 +40,7 @@
       default: {}
     }
 
-    if $facts['efi'] {
+    if fact('efi') {
```
bastelfreak, line 43: I would not change this. the `fact()` call is a function call which is slightly more expensive. even if `$facts['efi']` isn't set, it will return undef, i won't fail.

site/profile/manifests/core/cmms.pp
```diff
@@ -6,7 +6,7 @@
 ) {
   include profile::core::letsencrypt
 
-  $fqdn  = $facts[fqdn]
+  $fqdn  = fact('networking.fqdn')
```
bastelfreak, line 9: same here, and for the other occurrences. I wouldn't change it.

spec/classes/core/perfsonar_spec.rb
```diff
@@ -6,7 +6,7 @@
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let(:facts) { os_facts }
-      let(:fqdn) { facts[:fqdn] }
+      let(:fqdn) { facts[:networking]['fqdn'] }
```
bastelfreak: this let shouldn't be required because rspec-puppet-fact sets the fqdn properly.

jhoblitt: This might be a left over from the past... at one point `fqdn` (not a fact lookup) was used in some examples.

site/profile/manifests/ccs/nvidia.pp
jhoblitt: I'm aware it adds the overhead of a function call.  I prefer to use the dot notation as the same format works for both hiera and puppet code.  It also reduces the number of chars in a line and I find more compact code is easier to read.

Gemfile
jhoblitt: I swear that wasn't the case back when I originally added the lint gems. It looks like things have changed a bit:

```
    voxpupuli-puppet-lint-plugins (5.0.0)
      puppet-lint (~> 4.0)
      puppet-lint-absolute_classname-check (~> 4.0)
      puppet-lint-anchor-check (~> 2.0)
      puppet-lint-file_ensure-check (~> 2.0)
      puppet-lint-leading_zero-check (~> 2.0)
      puppet-lint-lookup_in_parameter-check (~> 2.0)
      puppet-lint-manifest_whitespace-check (~> 0.3, < 1.0.0)
      puppet-lint-optional_default-check (~> 2.0)
      puppet-lint-param-docs (~> 2.0)
      puppet-lint-param-types (~> 2.0)
      puppet-lint-params_empty_string-check (~> 2.0)
      puppet-lint-resource_reference_syntax (~> 2.0)
      puppet-lint-strict_indent-check (~> 3.0)
      puppet-lint-topscope-variable-check (~> 2.0)
      puppet-lint-trailing_comma-check (~> 2.0)
      puppet-lint-unquoted_string-check (~> 3.0)
      puppet-lint-variable_contains_upcase (~> 2.0)
      puppet-lint-version_comparison-check (~> 2.0)
```

I appears everything except `puppet-lint-package_ensure-check` can be removed.

spec/classes/core/perfsonar_spec.rb
jhoblitt: It wasn't needed anymore. It has been removed.

site/profile/manifests/core/cmms.pp
jhoblitt: :hankey:

Gemfile
```diff
@@ -4,20 +4,14 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
...
-  gem 'puppet-lint-strict_indent-check', '~> 2.1.0',           require: false
-  gem 'puppet-lint-unquoted_string-check', '~> 2.2.0',         require: false
-  gem 'puppet-lint-variable_contains_upcase', '~> 1.4.0',      require: false
-
-  gem 'facterdb', '< 2', require: false # 3.0.0 drops support for EL7
-
-  gem 'r10k',    require: false
-  gem 'toml-rb', require: false # puppet/telegraf
+  gem 'librarian-puppet',                             require: false
+  gem 'puppet-lint-package_ensure-check', '~> 0.2.0', require: false
```
bastelfreak, line 12: I asked in IRC/Slack if we should add it to voxpupuli-puppet-lint-plugins, I think that makes sense.

site/profile/manifests/core/selinux.pp
```diff
@@ -1,6 +1,6 @@
 class profile::core::selinux {
-  if ($facts['os']['selinux']['enabled']) {
-    notify { "os.selinux.enabled == ${facts['os']['selinux']['enabled']}":
+  if fact('os.selinux.enabled') {
+    notify { "os.selinux.enabled == ${fact('os.selinux.enabled')}":
```
tuxmea, line 3: Notify resource produces a change. If this is for information only, I would make use of the echo type.

.rubocop.yml
```diff
@@ -1,6 +1,7 @@
 ---
 require:
 - rubocop-rspec
```
bastelfreak: if you want, you can also use the rubocop config we ship in voxpupuli-test. That's designed for modules that are compatible with Puppet 7 and 8 (and I hope to move to 8 soon).

```yaml
---
inherit_gem:
  voxpupuli-test: rubocop.yml
```

spec/spec_helper.rb
```diff
@@ -137,6 +137,33 @@ def node_files
...
 
+# wrapper around #override_facts from voxpupuli-test that generates lsst's
+# default [networking] node facts.
+def lsst_override_facts(base_facts, **overrides)
+  override_facts(base_facts, **override_facts(gen_net_facts, **overrides))
+end
+
+# depends on node_params being set
+def gen_net_facts
+  if nodename(:host) =~ %r{\.lsst\.org$}
```
bastelfreak, line 138: I think you're never setting the `node` parameter in your spec files: https://github.com/puppetlabs/rspec-puppet/blob/main/lib/rspec-puppet/support.rb#L250-L255

I haven't tested it in your code base, but we use it in some vox pupuli modules and for some customers. But implementing it would mean to update all spec files that now use `lsst_override_facts()`, which is also some work. I don't think your current implementation is too bad, you could also just keep using it.

jhoblitt: I thought that for the `host` type, `:node` defaulted to the subject.

Looking through rspec-puppet, I'm not so sure it works that way:

https://github.com/rodjek/rspec-puppet/blob/c4ceecde20c9c53a388587dd37d6a2fc5dac50e5/lib/rspec-puppet/support.rb#L278-L279

Which might explain why the hierarchical networking facts weren't being set.

site/profile/manifests/core/selinux.pp
jhoblitt: The notify only triggers if selinux is enabled... and then the host is rebooted with it disabled and the notify will not trigger again.

spec/spec_helper.rb
bastelfreak: keep in mind that the gem is now released from the puppetlabs namespace:
https://github.com/puppetlabs/rspec-puppet/blob/c4ceecde20c9c53a388587dd37d6a2fc5dac50e5/lib/rspec-puppet/support.rb#L278-L279

FacterDB by default sets fqdn to foo.example.com. And when you set `node`, that will overwrite the hostname/domain/fqdn fact if I remember it correctly. At least we always set `node` and nothing else when we've puppet code that fqdn logic.

jhoblitt: Actually, it looks like the intent is that `:node` doesn't need to be set https://github.com/rodjek/rspec-puppet/blob/c4ceecde20c9c53a388587dd37d6a2fc5dac50e5/lib/rspec-puppet/support.rb#L271-L272 

So something isn't right...

Gemfile
jhoblitt: Sounds good. I'll remove the direct dep after the next voxpupuli-test release.

spec/spec_helper.rb
jhoblitt: `foo.example.com` isn't coming from rspec-puppet.  I think its coming from the facterdb data.  I also tried deleting the `:networking` key and setting `:node` but that results in `:networking` being `nil`.
```
From: /home/jhoblitt/github/lsst-control/spec/hosts/nodes/test_spec.rb:13 :

     8:       let(:facts) { os_facts }
     9: 
    10:       it { is_expected.to compile.with_all_deps }
    11:       it do
    12:         require 'pry'
 => 13:         binding.pry
    14:       end
    15:     end # on os
    16:   end # on_supported_os
    17: end

[1] pry(#<RSpec::ExampleGroups::TestCpLsstOrg::OnAlmalinux8X8664>)> facts[:networking]
=> {"domain"=>"example.com",
 "fqdn"=>"foo.example.com",
 "hostname"=>"foo",
 "interfaces"=>
  {"enp0s3"=>
    {"bindings"=>[{"address"=>"10.0.2.15", "netmask"=>"255.255.255.0", "network"=>"10.0.2.0"}],
     "bindings6"=>[{"address"=>"fe80::a00:27ff:feca:eec9", "netmask"=>"ffff:ffff:ffff:ffff::", "network"=>"fe80::", "scope6"=>"link", "flags"=>["permanent"]}],
     "duplex"=>"full",
     "ip"=>"10.0.2.15",
     "ip6"=>"fe80::a00:27ff:feca:eec9",
     "mac"=>"08:00:27:ca:ee:c9",
     "mtu"=>1500,
     "netmask"=>"255.255.255.0",
     "netmask6"=>"ffff:ffff:ffff:ffff::",
     "network"=>"10.0.2.0",
     "network6"=>"fe80::",
     "operational_state"=>"up",
     "physical"=>true,
     "scope6"=>"link",
     "speed"=>1000},
   "lo"=>
    {"bindings"=>[{"address"=>"127.0.0.1", "netmask"=>"255.0.0.0", "network"=>"127.0.0.0"}],
     "bindings6"=>[{"address"=>"::1", "netmask"=>"ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff", "network"=>"::1", "scope6"=>"host", "flags"=>["permanent"]}],
     "ip"=>"127.0.0.1",
     "ip6"=>"::1",
     "mtu"=>65536,
     "netmask"=>"255.0.0.0",
     "netmask6"=>"ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
     "network"=>"127.0.0.0",
     "network6"=>"::1",
     "operational_state"=>"unknown",
     "physical"=>false,
     "scope6"=>"host"}},
 "ip"=>"10.0.2.15",
 "ip6"=>"fe80::a00:27ff:feca:eec9",
 "mac"=>"08:00:27:ca:ee:c9",
 "mtu"=>1500,
 "netmask"=>"255.255.255.0",
 "netmask6"=>"ffff:ffff:ffff:ffff::",
 "network"=>"10.0.2.0",
 "network6"=>"fe80::",
 "primary"=>"enp0s3",
 "scope6"=>"link"}
```

```
From: /home/jhoblitt/github/lsst-control/spec/hosts/nodes/test2_spec.rb:14 :

     9:       let(:facts) { os_facts }
    10:       let(:node) { 'test2.cp.lsst.org' }
    11:       it { is_expected.to compile.with_all_deps }
    12:       it do
    13:         require 'pry'
 => 14:         binding.pry
    15:       end
    16:     end # on os
    17:   end # on_supported_os
    18: end

[1] pry(#<RSpec::ExampleGroups::Test2CpLsstOrg::OnAlmalinux8X8664>)> facts[:networking]
=> {"domain"=>"example.com",
 "fqdn"=>"foo.example.com",
 "hostname"=>"foo",
 "interfaces"=>
  {"enp0s3"=>
    {"bindings"=>[{"address"=>"10.0.2.15", "netmask"=>"255.255.255.0", "network"=>"10.0.2.0"}],
     "bindings6"=>[{"address"=>"fe80::a00:27ff:feca:eec9", "netmask"=>"ffff:ffff:ffff:ffff::", "network"=>"fe80::", "scope6"=>"link", "flags"=>["permanent"]}],
     "duplex"=>"full",
     "ip"=>"10.0.2.15",
     "ip6"=>"fe80::a00:27ff:feca:eec9",
     "mac"=>"08:00:27:ca:ee:c9",
     "mtu"=>1500,
     "netmask"=>"255.255.255.0",
     "netmask6"=>"ffff:ffff:ffff:ffff::",
     "network"=>"10.0.2.0",
     "network6"=>"fe80::",
     "operational_state"=>"up",
     "physical"=>true,
     "scope6"=>"link",
     "speed"=>1000},
   "lo"=>
    {"bindings"=>[{"address"=>"127.0.0.1", "netmask"=>"255.0.0.0", "network"=>"127.0.0.0"}],
     "bindings6"=>[{"address"=>"::1", "netmask"=>"ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff", "network"=>"::1", "scope6"=>"host", "flags"=>["permanent"]}],
     "ip"=>"127.0.0.1",
     "ip6"=>"::1",
     "mtu"=>65536,
     "netmask"=>"255.0.0.0",
     "netmask6"=>"ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
     "network"=>"127.0.0.0",
     "network6"=>"::1",
     "operational_state"=>"unknown",
     "physical"=>false,
     "scope6"=>"host"}},
 "ip"=>"10.0.2.15",
 "ip6"=>"fe80::a00:27ff:feca:eec9",
 "mac"=>"08:00:27:ca:ee:c9",
 "mtu"=>1500,
 "netmask"=>"255.255.255.0",
 "netmask6"=>"ffff:ffff:ffff:ffff::",
 "network"=>"10.0.2.0",
 "network6"=>"fe80::",
 "primary"=>"enp0s3",
 "scope6"=>"link"}
```

```
From: /home/jhoblitt/github/lsst-control/spec/hosts/nodes/test2_spec.rb:14 :

     9:       let(:facts) { os_facts.delete(:networking); os_facts }
    10:       let(:node) { 'test2.cp.lsst.org' }
    11:       it { is_expected.to compile.with_all_deps }
    12:       it do
    13:         require 'pry'
 => 14:         binding.pry
    15:       end
    16:     end # on os
    17:   end # on_supported_os
    18: end

[1] pry(#<RSpec::ExampleGroups::Test2CpLsstOrg::OnAlmalinux8X8664>)> facts[:networking]
=> nil
```

jhoblitt: When not using facterdb, `facts[:networking]` is missing without or without `:node` being set.

bastelfreak: yes foo.example.com is coming from facterdb. It's not required to delete the networking fact. can you try:

```ruby
let(:facts) { os_facts }
let(:node) { 'test2.cp.lsst.org' }
```

jhoblitt: I have convinced myself that this is an rspec-puppet bug (unless there is monkey patching going on somewhere).
```
[4] pry(#<RSpec::ExampleGroups::Test2CpLsstOrg::OnAlmalinux8X8664>)> facts_hash('test2.cp.lsst.org')[:networking]
=> nil
```

jhoblitt: I don't think there's anything that can be done in this site module other than the hack that's in the PR.

.rubocop.yml
jhoblitt: The vox config intentionally wasn't used because I disagree with with removing trailing commas, the `.` on line wrapped method calls being trailing instead of preceding, etc.

Let me see how bad it is to use it as a base with overrides.

spec/spec_helper.rb
bastelfreak: yes. I don't think the hack is bad.

I did a quick test and it works as expected in our regular modules: https://github.com/voxpupuli/puppet-example/pull/76/files

Maybe something is weird with the rspec setup in your case.

.rubocop.yml
bastelfreak: I also want the trailing commas back, but reintroducing at vox pupuli causes a lot of work. I hope I can combine that with the drop of ruby 2.7 soonish.

jhoblitt: It wasn't as bad as I thought it was going to be.

spec/classes/ccs/graphical_spec.rb
```diff
@@ -22,30 +22,30 @@
 
       it { is_expected.to compile.with_all_deps }
 
-      include_examples 'x2go packages', os_facts: os_facts
+      include_examples('x2go packages', os_facts:)
```
bastelfreak, line 25: oh is that valid syntax, `os_facts:`?

spec/spec_helper.rb
jhoblitt: > yes foo.example.com is coming from facterdb. It's not required to delete the networking fact. can you try:
> 
> ```ruby
> let(:facts) { os_facts }
> let(:node) { 'test2.cp.lsst.org' }
> ```

This was tested as the second of the 3 examples I posted. It does not work.

spec/classes/ccs/graphical_spec.rb
jhoblitt: That was a rubocop refactor when I raised the target ruby version.

spec/spec_helper.rb
jhoblitt: > Maybe something is weird with the rspec setup in your case.

I don't think there isn't anything out of the ordinary in the setup with the possible exception of calling `#add_stdlib_facts`.

bastelfreak (approved):

PR #1365, opened 2024-08-05
(Puppetfile) rm example42/network; EL7 only & uses legacy facts

jhoblitt:
This removes the example42/network module but leaves most of the hiera data in place for future conversion to NetworkManager configuration.


cbarria (approved):
Looks Great 👍🏻

PR #1385, opened 2024-08-08
(role/rke) bump DEV client to v1.5.12

cbarria:
bumping rke client version on dev to 1.5.12 to use k8s v1.28.12
[](https://github.com/rancher/rke/releases/tag/v1.5.12)

note: the releases no longer have the sha256 file of the binaries

### DIFF
--- a/spec/hosts/nodes/ayekan01.ls.lsst.org_spec.rb
+++ b/spec/hosts/nodes/ayekan01.ls.lsst.org_spec.rb
@@ -49,7 +49,7 @@ describe 'ayekan01.ls.lsst.org', :sitepp do
 
       it do
         is_expected.to contain_class('rke').with(
-          version: '1.5.10',
+          version: '1.5.12',
         )
       end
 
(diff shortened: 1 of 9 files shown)


spec/hosts/nodes/ayekan01.ls.lsst.org_spec.rb
dtapiacl, line 1: Why ayekan if its LS ?

dtapiacl (approved):
Looks good to me, but curious as to why Ayekan is being put on the list if its LS.

jhoblitt (approved):

spec/hosts/nodes/ayekan01.ls.lsst.org_spec.rb
cbarria: check the tests, according to that, is been tested as DEV. 
Ayekan was provisioned as LS but then moved to DEV and not reprovisioned, plus this is not the final place for the observability deployment I believe.

PR #1388, opened 2024-08-10
(role/perfsonar) add EL9 support

jhoblitt:
pS >= 5 is required for EL9


cbarria (approved):
Looks Great for now 👍🏻

PR #1395, opened 2024-08-14
(site/cp/cluster/lsstcam-ccs) enable mrtg

glennmorris:
At the request of the camera team in Chile, this enables mrtg monitoring of summit lsstcam hosts.
mrtg has previously been used for monitoring of CCS hosts. This is intended to be a temporary measure.


jhoblitt (approved):
OK...

PR #1403, opened 2024-08-19
(profile::core::lhnrouting) rm; was only used on el7

jhoblitt:
This profile was only used on el7 rke nodes, which have long since been upgraded to el9.


cbarria (approved):
that was the ghost interface to avoid an issue with EL7 right? either way, looks good. 👍🏻

jhoblitt:
It was a dummy interface to basically provide a hook for rhel legacy networking to inject routes.

PR #1405, opened 2024-08-19
(profile::core::hardware) enable amd_pstate driver on supported hw

jhoblitt:
```
[jhoblitt@elqui01 ~]$ cat /sys/devices/system/cpu/cpufreq/policy0/scaling_driver 
amd-pstate-epp
```


cbarria (approved):
Looks Great 👍🏻

PR #1406, opened 2024-08-20
(role/ccs-hcu-fb) new specialized ccs hcu role

glennmorris:


jhoblitt (commented):
Would it make sense to have a  -fb role since there are multiple hosts with the same services?

glennmorris:
> Would it make sense to have a -fb role since there are multiple hosts with the same services?

OK, I've redone it by adding a new role.

jhoblitt (approved):
This is acceptable but note that there is a catchall for roles which don't have a specific spec file.  This means every role added without a corresponding spec file will extended the length of the CI run since catchall doesn't parallelize.

glennmorris:
> This is acceptable but note that there is a catchall for roles which don't have a specific spec file. This means every role added without a corresponding spec file will extended the length of the CI run since catchall doesn't parallelize.

I've now added a basic spec file.

PR #1407, opened 2024-08-21
(cluster/lsstcam-ccs) send rsyslog to fluentbit ayekan

cbarria:
As requested, we need to send the rsyslogs to the dev instance of Fluentbit so the logs can reach OpenSearch, this is a temporary fix while the monitoring system is properly set at the Summit and all hosts change the endpoint to a deployment made for production.
I created a support spec for this in case we need to modify all the tests for the lsstcam-ccs cluster hierarchy. (even tho, it's temporary)


jhoblitt (approved):
LGTM.

PR #1411, opened 2024-08-22
(cluster/*) set ceph cluster tuned profile to latency-performance

jhoblitt:
It was mentioned in a Ceph Days talk that this may improve the consistentcy of ceph benchmarks.


cbarria (approved):
Looks Great 👍🏻

PR #1413, opened 2024-08-23
(role/hexrot) change versions and fix permissions

cbarria:
Te-Wei requested changes on this for the Comcam movement today.


dtapiacl (approved):
LGTM.

PR #1415, opened 2024-08-23
(common) disable pquota by default

jhoblitt:
This cmdline flag only applies to xfs.  It seems that pquota isn’t enabled
on /var/lib/docker on most (any?) k8s nodes, so it should be safe to remove
puppet enforcement on this cmdline flag.

This will only apply to the rke, rke2server, and rke2agent roles which
include profile::core::kernel::quota.  Once this flag has been removed
from cmdline on all nodes, this commit may be reverted and the ::pquota
profile removed from the listed roles.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -572,3 +572,5 @@ rke2::config:
   disable:
     - "rke2-ingress-nginx"
   disable-cloud-controller: true
+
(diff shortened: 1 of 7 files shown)


cbarria (approved):
Looks Great 👍🏻

PR #1417, opened 2024-08-26
(node/auxtel-mcm.cp, site/cp/role/ccs-mcm) configure ccs kafka brokers

glennmorris:
This requires https://github.com/lsst-it/lsst-puppet-hiera-private/pull/131


jhoblitt (approved):
As always, spec would be appreciated.

PR #1421, opened 2024-08-29
(role/comcam-fp) add tcp_tweak profile for s3daemon

cbarria:
These are the parameters that KT requested for the lsstcam-dc Nodes.

### DIFF
--- a/hieradata/role/ccs-dc.yaml
+++ b/hieradata/role/ccs-dc.yaml
@@ -7,6 +7,6 @@ classes:
   - "profile::core::common"
   - "profile::core::debugutils"
   - "profile::core::nfsclient"
-  - "profile::core::sysctl::lhn"
+  - "profile::core::sysctl::tcp_tweak"
 
 profile::ccs::common::sysctls: false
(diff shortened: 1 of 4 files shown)


dtapiacl (approved):
LGTM.

jhoblitt (commented):
Why is this different from the existing lhn profile?

cbarria:
> Why is this different from the existing lhn profile?

just 2 parameters are different (`net.ipv4.tcp_congestion_control` & `net.ipv4.tcp_slow_start_after_idle`)
you rather I keep the LHN and add another "profile" for these two parameters, or do I add this to the existing LHN profile?
@jhoblitt

csilva-cl:
> > Why is this different from the existing lhn profile?
> 
> just 2 parameters are different (`net.ipv4.tcp_congestion_control` & `net.ipv4.tcp_slow_start_after_idle`) you rather I keep the LHN and add another "profile" for these two parameters, or do I add this to the existing LHN profile? @jhoblitt

we should add these 2 parameters to the lhn profile, the rest is the same.

jhoblitt:
> just 2 parameters are different (`net.ipv4.tcp_congestion_control` & `net.ipv4.tcp_slow_start_after_idle`) you rather I keep the LHN and add another "profile" for these two parameters, or do I add this to the existing LHN profile? @jhoblitt

I'd rather add those sysctls to the existing LHN profile as support USDF transfers is the intended function of that profile.

PR #1430, opened 2024-09-05
(profile::core::sysctl::lhn) add new parameters

cbarria:
as requested :

- changed `net.ipv4.tcp_congestion_control` to `bbr`
- added `tcp_slow_start_after_idle` set to `0`


dtapiacl (approved):
LGTM.

PR #1431, opened 2024-09-05
(role/ccs-dc) add hosts entry for SLAC s3

cbarria:
same as the `comcam-fp01`

add entry on `/etc/hosts` for the SLAC s3 entry point


csilva-cl (approved):
looks good

PR #1449, opened 2024-09-09
(role/ccs-dc) add s3daemon instances

jhoblitt:
Add instances of the s3daemon to `ccs-dc` nodes.  The default port of `15555` was intentionally not used to avoid a potential conflict. The general scheme is to have an instance listening on port `15570` connected to the local site's s3/lfa instance, using the s3 credentials for buckets enumerated in https://rubinobs.atlassian.net/issues/IT-5301 (+ the addition of credentials for a `comcam` user following the same pattern).  A second instance listening on `15580` is connected to the appropriate usdf endpoint. Temporarily, there is a 3rd instance on the summit `lsstcam-dc*` nodes for acceptance testing of the production s3 endpoint.

Related to https://github.com/lsst-it/lsst-puppet-hiera-private/pull/135


hieradata/site/ls/cluster/lsstcam-ccs.yaml
```diff
@@ -7,3 +7,13 @@ ccs_software::service_email: "base-teststand-alerts-aaaaojwlqmczgqmavrwyopux5a@r
...
 ccs_monit::alert:
   - "base-teststand-alerts-aaaaojwlqmczgqmavrwyopux5a@rubin-obs.slack.com"
+
+s3daemon::instances:
+  ls-lsstcam:
+    s3_endpoint_url: "https://s3.ls.lsst.org"
+    port: 15570
+    image: "%{alias('s3daemon_default_image')}"
+  sdfembs3-lsstcam:
+    s3_endpoint_url: "https://sdfembs3.sdf.slac.stanford.edu"
```
ktlim: I'm not sure we can talk on the IPSec tunnels from the Base.  This may have to be s3dfrgw instead.

ktlim (approved):
I presume that `s3daemon_test` is temporary for the secrets.

tony-johnson (commented):
I am somewhat unsure what this pull request does. Note we are already using sdfembs3 on summit comcam (wthout s3daemon, although s3daemon is running and ready to be used) and summit lsstcam (using s3daemon). The JIRA linked, IT-5301 only discusses BTS/OODS, but these changes seem more pervasive than the ticket would suggest.

Can I get a summary of what this is intended to do?

womullan:
@tony-johnson we are to use s3daemon for oods as well instead of NFS. This is putting another s3daemon on each machine with a more local endpoint (like https://s3.cp.lsst.org) hence it is extensive. I am not clear on the hiera change  . But @jhoblitt should confirm.

hieradata/site/ls/cluster/lsstcam-ccs.yaml
womullan: correct - no route from base to ipsec tunnel

tony-johnson:
> @tony-johnson we are to use s3daemon for oods as well instead of NFS. This is putting another s3daemon on each machine with a more local endpoint (like https://s3.cp.lsst.org) hence it is extensive. I am not clear on the hiera change . But @jhoblitt should confirm.

I am OK with using s3daemon for oods, although I thought we had discussed this in the past and decided s3daemon provides little benefit for local (summit) connections and that we would use mc directly for s3/oods (this is what is currently set up for BTS).

But I am more concerned with the fact that I saw references to sdfembs3 in the pull request which do not seem relevant to oods. Also we already have a system for installing s3daemon on the summit machines using puppet, set up by @glennmorris and I want to make sure we are not ending up with 2 different ways of installing s3daemon. 

At a minimum this seems like it needs some more discussion.

hieradata/site/ls/cluster/lsstcam-ccs.yaml
jhoblitt: @ktlim I wasn't sure what to do for BTS. Obviously, need need a means of testing. Should this be changed to s3dfrgw? One option would be to configure another instance that connects to a local s3 endpoint but listens on a different port in order to allow testing with a "remote push" + "local push" workflow.

jhoblitt:
> I am not clear on the hiera change .

There's no change to the "live" hierarchy.  What was changed was the hierarchy used for unit tests, which was out of sync.

jhoblitt:
> I am OK with using s3daemon for oods, although I thought we had discussed this in the past and decided s3daemon provides little benefit for local (summit) connections and that we would use mc directly for s3/oods (this is what is currently set up for BTS).

My understanding was the summit local and USDF transfers were supposed to use the same mechanism in production as part of migration to the s3 the based oods.

> But I am more concerned with the fact that I saw references to sdfembs3 in the pull request which do not seem relevant to oods.

The oods will function the same at the summit and USDF.  Objects are pushed into an s3 endpoint, which in turn pushes a notification to a kafka topic, and the oods consumes those topics.

> Also we already have a system for installing s3daemon on the summit machines using puppet, set up by @glennmorris and I want to make sure we are not ending up with 2 different ways of installing s3daemon.

Chile IT long ago took an action item to roll out s3daemon and I announced that I was working on deploying it several months ago. The `profile::ccs::file_transfer` profile doesn't support multiple instances per node and is dependent on the system python version.  On el9 is system python is 3.9 while the dm baseline is already moving to 3.12.

womullan:
> am OK with using s3daemon for oods, although I thought we had discussed this in the past and decided s3daemon provides little benefit for local (summit) connections a

it is more that it gets us off NFS and then we should use the same mechanism.

womullan:
@glennmorris   if you and @jhoblitt can agree on this in short order Tony is ok with it (and I am) .. we would like to make sure we have one install method for s3daemon ..

glennmorris:
> @glennmorris if you and @jhoblitt can agree on this in short order Tony is ok with it (and I am) .. we would like to make sure we have one install method for s3daemon ..

No objection from me, so long as installing this will not affect the existing CCS usage of s3daemon.
I assume not, since the latter is referred to above, but please could we have confirmation?
Currently CCS starts an s3daemon systemctl service, and relies on being able to call ~ccs-ipa/s3daemon/git/python/s3daemon/send.py .

I'm sure this new approach is better than the existing CCS method and should replace it in the longer term.
Guidance on doing that would be appreciated.

jhoblitt:
> No objection from me, so long as installing this will not affect the existing CCS usage of s3daemon.
> I assume not, since the latter is referred to above, but please could we have confirmation?

This PR was designed not to impact the currently running s3daemon instances and to allow gradual transition.

glennmorris (approved):

hieradata/site/ls/cluster/lsstcam-ccs.yaml
jhoblitt: I'm going to remove sdfembs3 completely for `ls` for the moment. We will need to revisit how testing on BTS should work.

PR #1454, opened 2024-09-09
(site/dmz) retire unused dmz site & remove puppet/firewalld mod

jhoblitt:
Lets use only `puppetlabs/firewall` to manage iptables/nft instead of different mods within different roles.  In the future, we should evaluate migrating to `puppet/nftables`. `puppet/firewalld` may become an archived github repo in the near future.


dtapiacl (approved):
LGTM.

PR #1460, opened 2024-09-11
(role/laserrpi) add dco profile to laserrpi role

dtapiacl:
Ticket of reference: https://rubinobs.atlassian.net/browse/IT-5653


gseriche (approved):
LGTM

PR #1462, opened 2024-09-11
(profile::archive::common) rm unused services, accounts, & packages

jhoblitt:
The {comcam,auxtel}-archiver roles serve only as NFS servers (which should be retired before the start of operations).

This unused code was noticed while remove the usage of legacy facts in another feature branch.


cbarria (approved):
nice cleaning, looks great 👍🏻

PR #1473, opened 2024-09-18
(node/azar03.cp.lsst.org) setup as tmp lsstcam-archiver.cp.lsst.org

jhoblitt:
lsstcam-archiver.cp.lsst.org on esxi is being overwelmed by 10 dc nodes
and ssh connections are failing. @rhl & @a2 wanted to try a bare metal
node to see if it reduces the connection failures.


dtapiacl (approved):
LGTM.

dtapiacl:
> lsstcam-archiver.cp.lsst.org on esxi is being overwelmed by 10 dc nodes and ssh connections are failing. @rhl & @a2 wanted to try a bare metal node to see if it reduces the connection failures.

We recently updated the specs on this machine before we went out on vacation, the users reported problems even after the upgrade ?

jhoblitt:
> We recently updated the specs on this machine before we went out on vacation, the users reported problems even after the upgrade ?

Yes, the users were reporting connection failures and wanted to try a bare metal host.

PR #1480, opened 2024-09-27
(node/nfs3.cp.lsst.org) add comcam-archiver.cp.lsst.org export 

dtapiacl:


jhoblitt (changes requested):
Please update the spec examples.

jhoblitt (approved):
This looks good.

PR #1483, opened 2024-09-30
(node/comcam-archiver.cp) add EL9 network configuration & tests

dtapiacl:


hieradata/node/comcam-archiver.cp.lsst.org.yaml
```diff
@@ -1,105 +1,96 @@
...
+        uuid: "b28d29b1-83cc-4df2-9c65-22cd9d97d0af"
+        type: "ethernet"
+        interface-name: "eno1"
+      ethernet: {}
+      ipv4:
+        method: "auto"
+      ipv6:
+        method: "disabled"
+      proxy: {}
+  eno2:  #dds interface
```
jhoblitt: OSPL/DDS no longer runs on the -archiver hosts. I think we should disable the dds interface and free any route53 records.

jhoblitt (changes requested):
This looks good but as we are revisiting the network configuration it makes sense to get rid of the dds interface since it hasn't been used for more than a year.

cbarria (approved):
had to put `nfs::client_enabled: true` to enable it. Host was re-provision on this branch.

jhoblitt (approved):

gseriche (approved):
lgtm

PR #1488, opened 2024-10-04
(node/rucio01.ls.lsst.org) deploy rucio

dtapiacl:
Requirements per ticket information on IT-5509:

• Install the following rpms, xrootd, xrootd-selinux, xrootd-libs, xrootd-client, xrootd-client-libs, xrootd-server-libs, xrootd-server
• modify the /lib/systemd/system/[xrootd@.service](mailto:xrootd@.service) and /lib/systemd/system/[cmsd@.service](mailto:cmsd@.service) and change the User and Group inside these two unit files to saluser and it group (current it is user xrootd)
• Change the owner of directory /etc/xrootd, /var/{log, run, spool}/xrootd to saluser (this probably need to be done every time there is an update of the xrootd rpms)
• Allow user bwhite and wyang to do a) sudo su - saluser and b) sudo systemctl restart xrootd@summit. We will add Sajid Ali to this list of users once he has an account
• Place VM in BTS-LHN network.
• Mount the filesystems KT mentioned.

### DIFF
--- a/hieradata/role/rucio.yaml
+++ b/hieradata/role/rucio.yaml
@@ -1,12 +1,32 @@
 ---
 classes:
   - "profile::core::common"
-  - "profile::core::debugutils"
+  - "profile::core::firewall"
+  - "profile::core::nfsclient"
   - "profile::core::rucio"
-  - "python"
 
-python::version: "python36"
-python::pip: "present"
-python::dev: "present"
-python::virtualenv: "present"
-profile::core::common::manage_firewall: false
+profile::core::firewall::firewall:
+  "254 accept incoming on port 1094 from SLAC subnet":
+    proto: "tcp"
+    state: "NEW"
+    dport: "1094"
+    source: "134.79.23.0/24"
+    jump: "accept"
+  "255 accept outgoing on port 1094 to SLAC subnet":
+    proto: "tcp"
+    state: "NEW"
+    sport: "1094"
+    destination: "134.79.23.0/24"
+    jump: "accept"
+  "256 accept incoming on port 1095 from SLAC subnet":
+    proto: "tcp"
+    state: "NEW"
+    dport: "1095"
+    source: "134.79.23.0/24"
+    jump: "accept"
+  "257 accept outgoing on port 1095 to SLAC subnet":
+    proto: "tcp"
+    state: "NEW"
+    sport: "1095"
+    destination: "134.79.23.0/24"
+    jump: "accept"
--- a/site/profile/manifests/core/rucio.pp
+++ b/site/profile/manifests/core/rucio.pp
@@ -2,6 +2,20 @@
 #   Install required rucio packages
 #
 class profile::core::rucio {
+  include profile::core::letsencrypt
+
+  #  Host FQDN
+  $fqdn = $facts['networking']['fqdn']
+
+  #  Signed Certificate Location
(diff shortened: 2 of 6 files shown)


hieradata/role/rucio.yaml
```diff
@@ -1,12 +1,5 @@
...
-  - "profile::core::debugutils"
+  - "profile::core::nfsclient"
   - "profile::core::rucio"
-  - "python"
-
-python::version: "python36"
-python::pip: "present"
-python::dev: "present"
-python::virtualenv: "present"
-profile::core::common::manage_firewall: false
```
jhoblitt, line 12: I would like all new or heavily updated rules for here forward to enable iptables/nft rules.  I have already started doing this. See the puppetdb role for a good example of this: https://github.com/lsst-it/lsst-control/blob/3b2a02929b41363945b8ce935255b0b15d104d52/hieradata/role/puppetdb.yaml#L28-L48

site/profile/manifests/core/rucio.pp
```diff
@@ -1,63 +1,38 @@
...
+    ensure => 'installed',
   }
-
-  #  Install Pip3 Packages
-  package { $pip_packages:
-    ensure   => 'present',
-    provider => 'pip3',
+  file { [
+      '/lib/systemd/system/xrootd@.service',
+      '/lib/systemd/system/cmsd@.service',
```
jhoblitt, line 19: Why would saluser need to be able to edit the service units?

site/profile/manifests/core/rucio.pp
```diff
@@ -1,63 +1,38 @@
...
+    ]:
+      ensure => file,
+      mode   => '0644',
+      owner  => 'saluser',
```
jhoblitt, line 23: Why is `saluser` being used as the role user?  This isn't a tssw related service and it doesn't interact with sal/dds.

site/profile/manifests/core/rucio.pp
jhoblitt: Are the service units being installed by a package?  If so, the file resource(s) need to be ordered on the package.

jhoblitt (changes requested):

site/profile/manifests/core/rucio.pp
dtapiacl: They want to use `saluser` as a shared account to manage the service.

site/profile/manifests/core/rucio.pp
dtapiacl: They want to use `saluser` as a service account to manage the xrootd related services.

jhoblitt: Changing the ownership of these files to `saluser` does not allow the `saluser` to start or stop those services. Nor does it grant `saluser` the ability to trigger systemd to re-read those files after they have changed.

site/profile/manifests/core/rucio.pp
dtapiacl: They are using `saluser` as a service account to manage the xrootd related services. 

These service units are installed by the xrootd package.

site/profile/manifests/core/rucio.pp
jhoblitt: I also object to role accounts editing service units.  Changes should be made via puppet.

PR #1491, opened 2024-10-07
(role/rke) bump LS client to v1.6.2

cbarria:
this is October upgrade on the Base... `rke client` is bump to `v1.6.2` and `k8s` to `v1.29.8`

this PR goes incredible well with https://github.com/lsst-it/k8s-cookbook/pull/622


dtapiacl (approved):
LGTM.

PR #1499, opened 2024-10-11
(Puppetfile) fix clustershell sha

glennmorris:
The branch was rebased, and the old sha no longer exists.


jhoblitt (approved):
When we fork an upstream module, we should create either a tag or a branch for modifications that is intended to be static, that way this doesn't happen while trying to get changes upstreamed.

PR #1509, opened 2024-10-22
(profile::ccs::file_transfer) default s3daemon to true

glennmorris:
Also remove per-host settings that now match the default.


tony-johnson (approved):
Looks good to me, unless we want to include additional changes discussed in SLACK in the same update

PR #1513, opened 2024-10-23
(cluster/auxtel-ccs) switch to java17

glennmorris:
For deployment after reinstallation with EL9.


tony-johnson (approved):
Looks good to me.

PR #1515, opened 2024-10-24
(*) change the priority variable from a String to an integer

badenerb:


jhoblitt (changes requested):
Nice work on finding the issue.

The one quibble I have is that the PR title/commit message subject line should be imperative instead of using past tense.

https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form

jhoblitt (approved):
Looks good now.

PR #1519, opened 2024-10-28
(common/rke2) bump rke2 to v1.29.9 global

cbarria:
During the last summit upgrade i forgot to update `Elqui`, so I'm bumping all sites version to `v1.29.9` (instead of just `DEV`)


jhoblitt (approved):
LGTM.

PR #1520, opened 2024-11-01
(cluster/elqui) misc tuning for ceph rgw

jhoblitt:


hieradata/cluster/elqui.yaml
```diff
@@ -80,6 +80,7 @@ nm::connections:
       bond:
         miimon: "100"
         mode: "802.3ad"
+        xmit_hash_policy: "layer3+4"
```
dtapiacl, line 83: Network + Transport layer ?

dtapiacl (approved):
LGTM.

hieradata/cluster/elqui.yaml
jhoblitt: IP address + port

PR #1524, opened 2024-11-04
(role/daq-mgt) Update DAQ to R5-V12.1

jgthayer:
Update DAQ to R5-V12.1


jgthayer:
@jhoblitt Once, long ago, I was able to "self review" these. I have lost that ability. I'd be happy to get it back, particularly for trivial changes like this.

glennmorris (approved):

PR #1525, opened 2024-11-04
(node/auxtel-fp01.cp) add el9 network config & tests

dtapiacl:
Convert network configuration from EL7 to EL9 in preparation for its OS upgrade.

### DIFF
--- a/hieradata/node/auxtel-fp01.cp.lsst.org.yaml
+++ b/hieradata/node/auxtel-fp01.cp.lsst.org.yaml
@@ -1,37 +1,77 @@
 ---
-network::interfaces_hash:
-  em2:  # dds-auxtel
-    bootproto: "dhcp"
-    defroute: "no"
-    nozeroconf: "yes"
-    onboot: "yes"
-    type: "Ethernet"
-  em3:  # not connected
-    bootproto: "none"
-    onboot: "no"
-    type: "Ethernet"
-  em4:  # fqdn
-    bootproto: "dhcp"
-    defroute: "yes"
-    onboot: "yes"
-    type: "Ethernet"
-
-network::mroutes_hash:
-  em2:
-    routes:
-      "139.229.147.0/24": "139.229.170.254"
-      "139.229.166.0/24": "139.229.170.254"
-      "139.229.167.0/24": "139.229.170.254"
-      #"139.229.170.0/24": "139.229.170.254"
-      "139.229.178.0/24": "139.229.170.254"
+nm::connections:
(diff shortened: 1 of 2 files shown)


cbarria (approved):
Looks Great 👍🏻

hieradata/node/auxtel-fp01.cp.lsst.org.yaml
```diff
@@ -1,37 +1,85 @@
...
+        uuid: "17b7e246-327c-76fb-6648-b59ca1bdc81e"
+        type: "ethernet"
+        interface-name: "eno1"
+      ethernet: {}
+      ipv4:
+        method: "auto"
+      ipv6:
+        method: "disabled"
+      proxy: {}
+  eno2:  # dds-auxtel
```
jhoblitt: I'm unsure if CSCs are still running on auxtel-fp01.  I believe the intent was the would only be running on auxtel-mcm.

This seems OK as is but I would ask if the dds interface configuration can be removed.

gseriche (approved):
lgtm

hieradata/node/auxtel-fp01.cp.lsst.org.yaml
dtapiacl: According to Tony they have never used it. I will proceed to clean that up.

https://rubin-obs.slack.com/archives/C07QM70QAR2/p1730829490064959

jhoblitt (changes requested):
The spec file name doesn't need to include `.yaml`. E.g. `auxtel-fp01.cp.lsst.org.yaml_spec.rb`

dtapiacl:
> The spec file name doesn't need to include `.yaml`. E.g. `auxtel-fp01.cp.lsst.org.yaml_spec.rb`

You are correct, this was a typo on my end that i did not see. Thank you !

hieradata/node/auxtel-fp01.cp.lsst.org.yaml
```diff
@@ -51,7 +91,7 @@ nfs::client_enabled: true
 nfs::client_mounts:
   /net/self/data:
     share: "data"
-    server: "%{facts.networking.fqdn}"
+    server: "auxtel-fp01.cp.lsst.org"
```
jhoblitt: Why does this need to be changed?

dtapiacl: No particular reason, i have updated the code.

jhoblitt (approved):

PR #1526, opened 2024-11-05
(cluster/yagan) add CNI static plugin

cbarria:
in order to put a static IP on the `multus` interface, `static` plugin must be installed on `RKE1` (comes out-of-the-box on RKE2, migration is recommended)

### DIFF
--- a/hieradata/cluster/yagan.yaml
+++ b/hieradata/cluster/yagan.yaml
@@ -4,3 +4,4 @@ clustershell::groupmembers:
 profile::core::k8snode::enable_dhcp: true
 profile::core::ospl::enable_rundir: true
 tuned::active_profile: "latency-performance"
+cni::plugins::enable: ["macvlan", "static"]
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -104,6 +104,9 @@ lookup_options:
   rke2::config:
     merge:
       strategy: "deep"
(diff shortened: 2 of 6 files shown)


hieradata/common.yaml
```diff
@@ -104,6 +104,9 @@ lookup_options:
   rke2::config:
     merge:
       strategy: "deep"
+  cni::plugins::enable:
+    merge:
+      strategy: "deep"
```
dtapiacl, line 109: What does the merge strategy line of code do ?

hieradata/cluster/yagan.yaml
dtapiacl, line 1: This looks good to me, good to know that you just cant add a piece of static code to a multus interface without first enabling the plugin.

dtapiacl (approved):
LGTM.

hieradata/common.yaml
cbarria: this means that if puppet finds more config on "that" on a different level of the hierarchy, it will not overwrite, but combine all the configuration together.

hieradata/cluster/yagan.yaml
cbarria: but in order to add the static code in the ipam section it needs to have the plugin installed.. its a binary on `/opt/cni/bin`

jhoblitt (approved):
LGTM.

PR #1530, opened 2024-11-07
(node/foreman.cp) increase vlan1201|CDS-ARCH-DDS dhcp pool to 150 ips

jhoblitt:
There are current 50 pods running on yagan with an IP from 139.229.166/24 and there are complaints about notebook startup failing as the DHCP pool is exhausted.


badenerb (approved):
LGTM!!

PR #1543, opened 2024-11-19
(node/observer01.cp.lsst.org) provision host as EL9 & add test

badenerb:


dtapiacl (approved):
LGTM.

jhoblitt (commented):
If this host is on DHCP, do we need a node layer for it?

PR #1550, opened 2024-11-27
(profile::pi::add_usb.pp) add serialusb class add class to role add test

badenerb:


jhoblitt (commented):
I won't block this being merged without tests as I'm going to be away for the next few days... but I recommend attempting to write some tests. Perhaps @cbarria could help this week?

cbarria (approved):
Looks Great 👍🏻

PR #1554, opened 2024-12-02
(role/hexrot) fix nexus packages for hexrot

cbarria:
the `ctio nexus` is no longer available, so Te-Wei moved his `Hexrot` Packages into [repo-nexus.lsst.org](repo-nexus.lsst.org) and removed the used of `runM2Cntlr` as it is now managed by the `ts_m2gui` conda packages now.


dtapiacl (approved):
LGTM

PR #1561, opened 2024-12-05
(role/auxtelmcm) add ccs_database

dtapiacl:
This PR cleans up the old db puppet profile assigned to this role  to use instead `ccs_dabasse` puppet module.


badenerb (approved):
LGTM! 🍍

PR #1586, opened 2024-12-18
(node/auxtel-dc01.cp) add el9 network config & tests

dtapiacl:


gseriche (approved):
no issues from my side. gg

PR #1609, opened 2025-01-06
(node/auxtel-fp01.cp) add lhn interface

dtapiacl:
Enables eno3 to serve as lhn interface for auxtel-fp01.cp.lsst.org. @tony-johnson agreed to use the 1G interface as they will only push small size files.


julioconstanzo (approved):
LGTM

badenerb (approved):
LGTM!

PR #1611, opened 2025-01-13
(role/laserrpi) add new udev rule ttyusb

cbarria:
this rule replaces the one with the fix vendor ID for a more generic rule that can be more managed.


jhoblitt (approved):
LGTM.  Is the intent to fully replace the `profile::pi::ftdi` profile class?

cbarria:
> LGTM.  Is the intent to fully replace the `profile::pi::ftdi` profile class?

Yes, i will deprecate the ftdi profile and propagate this changes to all rpi once it is ok 👍🏻

PR #1617, opened 2025-01-27
(node/love01.ls) retire love ls from infrastructure

dtapiacl:
According to Michael, love01.ls.lsst.org will no longer be in use. This PR cleans up any configuration pertaining to this specific node.


dtapiacl:
Node was removed from foreman.ls.lsst.org.

jhoblitt (approved):

jhoblitt:
> Node was removed from foreman.ls.lsst.org.

What is the name of the node now?  We need to keep a host entry in foreman so that the bmc remains accessible.

PR #1619, opened 2025-01-28
(Puppetfile) update java_artisanal

glennmorris:
This version ensures symlinks for java in /usr/bin (to /etc/alternatives) do not get destroyed.
Some version of this is needed (on cp_production) before the next summit OS updates (Thursday?).


dtapiacl (approved):
LGTM.

PR #1622, opened 2025-01-29
(site/cp) switch CCS SAL hosts from DDS to Kafka

glennmorris:
This requires https://github.com/lsst-it/lsst-puppet-hiera-private/pull/144
This should come after https://github.com/lsst-it/lsst-control/pull/1571

### DIFF
--- a/hieradata/site/cp.yaml
+++ b/hieradata/site/cp.yaml
@@ -52,6 +52,11 @@ accounts::user_list:
 profile::core::common::disable_ipv6: true
 profile::core::docker::version: "24.0.9"
 
+ccs_sal::dds: false
+ccs_sal::kafka: true
+ccs_sal::kafka_broker_address: "sasquatch-summit-kafka-bootstrap.lsst.codes:9094"
+ccs_sal::kafka_registry_url: "https://summit-lsp.lsst.codes/schema-registry"
+
 ccs_sal::lfa:
   id: "%{alias('ccs_sal::lfa::id')}"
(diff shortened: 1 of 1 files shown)


tony-johnson (dismissed):
Right now TTS (.tu) is still running DDS. It is fine to put in the new sal::kafka settings, but I am not sure of the effect of:

ccs_sal::dds: false

which might be premature.

glennmorris:
> Right now TTS (.tu) is still running DDS. It is fine to put in the new sal::kafka settings, but I am not sure of the effect of:
> 
> ccs_sal::dds: false
> 
> which might be premature.

Oh, then I had misunderstood what was wanted here.
So you just want .cp switching for now?

tony-johnson:
> > Right now TTS (.tu) is still running DDS. It is fine to put in the new sal::kafka settings, but I am not sure of the effect of:
> > ccs_sal::dds: false
> > which might be premature.
> 
> Oh, then I had misunderstood what was wanted here. So you just want .cp switching for now?

More precisely, it is fine to install the latest version of /etc/ccs/setup-sal-kafka everywhere including TTS, but on TTS we are still running DDS so we should not remove anything needed by DDS, nor should we change the logic in /etc/ccs/setup-sal (which defaults to using DDS). I am just not sure of the effect of

ccs_sal::dds: false

but if it disables DDS then we don't want it on TTS quite yet.

glennmorris:
@tony-johnson I have now changed this to only affect site/cp.

jhoblitt (approved):

PR #1624, opened 2025-02-05
(kueyen) Add more nodes. Add a variant for new nodes.

gseriche:
Add more nodes for Kueyen


dtapiacl (changes requested):
This looks good, you are only missing the variant tests.

dtapiacl (approved):
Looks good to me. Thanks for creating the variants. Make sure CI Passes.

PR #1625, opened 2025-02-05
(role/rke2*) disable rke2-snapshot-controller related helm charts

jhoblitt:
As the chart rke2-snapshot-controller-crd contains an out of date crd causing problems with rook and we aren't using this functionality.


cbarria (approved):
Thanks so much! Looks Great 👍🏻

PR #1642, opened 2025-02-12
(node/lsstcam-{fcs,quad,rebps,vac}) set ccs_hcu quadbox flag

glennmorris:
This just controls whether a utility script gets installed.


gseriche (approved):
lgtm

PR #1649, opened 2025-02-21
(pukem) provision cluster with rke2

csilva-cl:


hieradata/cluster/pukem.yaml
```diff
@@ -0,0 +1,6 @@
+---
+clustershell::groupmembers:
+  pukem: {group: "pukem", member: "pukem[01-04]"}
+profile::core::ospl::enable_rundir: true
```
cbarria: this is for the OSPL daemon not really needed here

hieradata/cluster/pukem.yaml
```diff
@@ -0,0 +1,6 @@
...
+profile::core::k8snode::enable_dhcp: true
```
cbarria, line 4: this is just in case the cluster uses Multus or any CNI plugin (macvlan, etc.)

cbarria (changes requested):
only those 2 changes, also might need another pukem node test for rke2agent and variant c6420.

hieradata/cluster/pukem.yaml
csilva-cl: correct

hieradata/cluster/pukem.yaml
csilva-cl: good catch, removed.

cbarria (approved):
Looks Great 👍🏻

PR #1650, opened 2025-02-27
(node/auxtel-ill-control.cp) add missing nfs mount

dtapiacl:


jhoblitt (changes requested):
I suspect CI is failing because the nfs client profile isn't part of the fiberspec role.

gseriche (approved):
lgtm

jhoblitt (approved):

PR #1657, opened 2025-03-12
(role/hexrot) add/update conda packages

cbarria:
- added `conda` packages: `ts-rotgui`, `ts-hexgui`, `ts-mtdomegui`, `ts-guitools`
- set permissions for their logs and python folder
- creates the symblinks for the binaries
- updates versions for ts-mt2gui (conda) and ts_config_mttcs (repo)
- added quick profile to enable xfce and set GUI by default
- got NI Lab View runtime into the local Nexus, pulling from there. (ni_packages)

### DIFF
--- a/hieradata/role/hexrot.yaml
+++ b/hieradata/role/hexrot.yaml
@@ -10,6 +10,11 @@ classes:
   - "profile::core::x2go"
   - "profile::ts::hexrot"
   - "profile::ts::nexusctio"
+  - "profile::util::xfce"
+
+packages:
+  - "freeglut-devel"
+
 files:
   /rubin:
     ensure: "directory"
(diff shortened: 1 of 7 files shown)


gseriche (approved):
LGTM, it seems a lot of work to make it work.

jhoblitt (approved):
This looks reasonable.  I'm not sure that the changeset/PR title are as helpful as they could be.  I suspect the main purpose of this change is to update `ts_config_mttcs.git`?

cbarria:
> This looks reasonable. I'm not sure that the changeset/PR title are as helpful as they could be. I suspect the main purpose of this change is to update `ts_config_mttcs.git`?

yeah, sorry for the title, the most important part is to add new packages, update older ones and get new GUIs... software is trying to remove the NI packages by migrating the GUIs into python, so, in the future we will probably remove the `ni_packages` profile.

PR #1658, opened 2025-03-17
(role/daq-mgt) Update DAQ to R5-V13.3

jgthayer:
Update Camera to DAQ R5-13.3. Note that this also requires the setting of the sysctl parameter kernel.sched_rt_runtime_us=-1 to allow the gds_relay to change CPUSchedulingPolicy and CPUSchedulingPriority.

### DIFF
--- a/hieradata/role/daq-mgt.yaml
+++ b/hieradata/role/daq-mgt.yaml
@@ -24,7 +24,7 @@ chrony::queryhosts:
   - "192.168/16"
 chrony::clientlog: true
 
-daq::daqsdk::version: "R5-V13.2"
+daq::daqsdk::version: "R5-V13.3"
 daq::rptsdk::version: "V3.5.3"
 
 # dhcp for DAQ + clients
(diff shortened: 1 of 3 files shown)


glennmorris (commented):
The version change looks fine of course. Wrt to the sysctl change, I see from
https://access.redhat.com/solutions/1604

"In short, setting sched_rt_runtime_us to -1 can be extremely dangerous. A value of -1 means no limit. In other words, a "run-away" real-time task will be permitted to monopolise a CPU which could (potentially) lock up a system."

So I don't know if this is something that needs discussion, eg in Slack somewhere.

jgthayer:
Yes I saw that too, but if I want to reliably meet the Guider latency requirement, I need the relay to operate at rt priority. There is little chance of it overwhelming the rest of the daq-management machine because it is very bursty.

Gregg

jhoblitt (approved):
I'm OK with giving the `sched_rt_runtime_us` value a try as it only applies to hosts which @jgthayer is exclusively responsible for.

PR #1659, opened 2025-03-17
(role/fiberspec.yaml) add docker profile

dtapiacl:
Ticket of reference: https://rubinobs.atlassian.net/browse/IT-6015


cbarria (approved):
Looks Great 👍🏻

PR #1660, opened 2025-03-19
(common/firewall) add antu and set port for node_exporter

cbarria:
- adds the `ipset` entry for `antu`.
- replaces `ayekan` with `antu` on the `firewall rule`.
- sets firewall rule for dev on port `9100`


jhoblitt (commented):
Could we allow ruka too?

cbarria:
> Could we allow ruka too?

Added the `DEV` segment where `Ruka` is 👍🏻

csilva-cl (approved):
looks good, thanks

PR #1668, opened 2025-03-26
(role/rke2) set dhcp to service_only on multus clusters

cbarria:
this should solve the version problem between the dhcp and macvlan (using the defaults in rke2)


jhoblitt (approved):
Thank you for cleaning up after me... I'm OK with this being merged once the specs are updated / CI is passing.

PR #1697, opened 2025-04-09
(luan) RKE1 -> RKE2 migration

dtapiacl:


hieradata/cluster/luan.yaml
```diff
@@ -1,4 +1,46 @@
 ---
 clustershell::groupmembers:
   luan: {group: "luan", member: "luan[01-05]"}
+profile::core::ospl::enable_rundir: true
```
jhoblitt: This is only needed to support DDS.  We had never needed it for luan and we should be removing it from all clusters shortly.

hieradata/cluster/luan.yaml
```diff
@@ -1,4 +1,46 @@
...
+profile::core::k8snode::enable_dhcp: true
```
jhoblitt: This is only used for multus. We don't want it on luan.  The cni-dhcp service will need to be manually removed from nodes.

hieradata/cluster/luan.yaml
```diff
@@ -1,4 +1,46 @@
...
+  eno1np0:
+    content:
+      connection:
+        id: "eno1np0"
+        uuid: "8d5f0c26-a6c6-3704-92a5-c347e9fbbf28"
+        type: "ethernet"
+        interface-name: "eno1np0"
+      ethernet: {}
+      ipv4:
+        method: "auto"
```
jhoblitt, line 16: We should probably be using bridge interfaces for all baremetal clusters for consistency but as this is a rush, it is OK to use the physical interface this time.  We should plan to rebuild this cluster again and convert to using LAGs.

jhoblitt (changes requested):

gseriche (approved):
lgtm

jhoblitt (approved):
We should plan to remove the agent role and convert to LAGs in the near-ish future.

PR #1704, opened 2025-04-16
(role/hexrot) mayor bump python 3.12 and conda packages

cbarria:
- DM went to python 3.12.
- bump most packages version.
- added few more packages (re order everything in alphabetical order, including tests, because it was making mad.)
- changed behavior with repo version, now is set on yaml manifest and pass to the class.


dtapiacl (approved):
Looks great. Thos are a lot of version bumps.

gseriche (approved):
lgtm

PR #1707, opened 2025-04-16
(openvpn) openvpn install with lets encrypt

dtapiacl:
Manages install of openvpn with lets encrypt and version control.

This PR goes well with: https://github.com/lsst-it/k8s-cookbook/pull/959

### DIFF
--- a/hieradata/role/openvpn.yaml
+++ b/hieradata/role/openvpn.yaml
@@ -1,0 +1,7 @@
+---
+classes:
+  - "profile::core::common"
+  - "profile::core::openvpn"
+
+profile::core::openvpn::version: "3.0.0_2b84043e"
+profile::core::openvpn::cluster: "vpn.%{::site}.lsst.org"
--- a/site/profile/manifests/core/openvpn.pp
+++ b/site/profile/manifests/core/openvpn.pp
@@ -1,0 +1,86 @@
+# @summary
+#   Installs openvpn connect server.
+#
+# @param version
+#   Sets version lock for openvpn package.
+#
+# @param cluster
+#   Generates an additional certificate for vpn round robin setup.
+#
+class profile::core::openvpn (
+  String[1] $version,
+  String[1] $cluster,
+) {
+  include yum::plugin::versionlock
+  include profile::core::letsencrypt
+
+  yumrepo { 'as-repo-rhel9':
+    ensure   => 'present',
+    name     => 'openvpn-access-server',
+    descr    => 'OpenVPN Access Server',
+    baseurl  => 'http://as-repository.openvpn.net/as/yum/rhel9/',
+    gpgkey   => 'https://as-repository.openvpn.net/as-repo-public.gpg',
+    gpgcheck => '1',
+    enabled  => '1',
+  }
+
+  package { 'openvpn-as':
+    ensure  => $version,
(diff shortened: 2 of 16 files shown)


hieradata/role/openvpn.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
jhoblitt, line 1: There should be spec coverage for the role.

spec/classes/core/openvpn_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'profile::core::openvpn' do
+  on_supported_os.each do |os, os_facts|
+    next unless os =~ %r{almalinux-9-x86_64}
```
jhoblitt, line 7: There should be spec coverage of the letsencrypt resources.

Ideally, the profile tests would be does as a `shared_example`, that can also be included in the role spec file.

site/profile/manifests/core/openvpn.pp
```diff
@@ -0,0 +1,86 @@
...
+
+  file { '/usr/local/openvpn_as/etc/web-ssl/ca.crt':
+    ensure  => 'link',
+    target  => "${le_root}/fullchain.pem",
+    force   => true,
+    require => Letsencrypt::Certonly[$fqdn],
+  }
+
+  # Restart OpenVPN Access Server.
+  exec { 'restart_openvpn_service':
```
jhoblitt: I don't see a notification being sent to this resource -- is it needed?

jhoblitt (changes requested):
This appears to install the openvpn packages.  How is the service configured? Do we need to manage the service state?

site/profile/manifests/core/openvpn.pp
dtapiacl: No this is not needed, i updated this to ensure that service is always running.

dtapiacl:
> This appears to install the openvpn packages. How is the service configured? Do we need to manage the service state?

I have updated the code to ensure that the service is always running and is enabled.

jhoblitt (commented):
In this repo, the `()` value is supposed to refer to the hiera layer or puppet class being modified.

What is the relationship between this PR and https://github.com/lsst-it/k8s-cookbook/pull/917?

dtapiacl:
> In this repo, the `()` value is supposed to refer to the hiera layer or puppet class being modified.
> 
> What is the relationship between this PR and [lsst-it/k8s-cookbook#917](https://github.com/lsst-it/k8s-cookbook/pull/917)?

This PR will handle the installation of the ovpn package with Let's Encrypt while giving us control of the versions installed.  This PR relates directly to https://github.com/lsst-it/k8s-cookbook/pull/959 and it supersedes (917).  

PR (959) deploys the OVPN database with its operator in a galera setup. This is currently being tested in Ayekan and is working as expected at vpn.dev.lsst.org, and I have not run into any issues while linking the ovpn server to the Kubernetes database, i also updated the server from version 2.14 to 3.0 without problems.

jhoblitt (commented):
What does this profile do that the a forge module such as https://forge.puppet.com/modules/puppet/openvpn/readme doesn't do?

PR #1723, opened 2025-04-23
(Puppetfile) Switch Docker to official version

badenerb:


badenerb:
The fix was merged here: https://github.com/lsst-it/lsst-control/pull/1725

PR #1736, opened 2025-05-05
(role/rke/rke2) bump CP rke client v1.7.6 and rke2 k8s v1.31.7

dtapiacl:


hieradata/site/ls/role/rke.yaml
```diff
@@ -1,3 +1,2 @@
 ---
 profile::core::docker::version: "24.0.9"
```
jhoblitt: I think this entire file can go as `profile::core::docker::version: "24.0.9"` is set in `profile::core::docker::version`. I suspect we can bump the docker version in `site/profile/data/common.yaml` and remove it from `site/<site>.yaml` completely (in another PR).

jhoblitt (changes requested):

hieradata/site/ls/role/rke.yaml
dtapiacl: I have opened a PR to bump docker version in common. Im not sure what you mean by `remove it from site/<site>.yaml` can you point me to this ?

jhoblitt (approved):
LGTM.

PR #1744, opened 2025-05-08
(role/comcam-archiver) rm

jhoblitt:
This retires the comcam-archiver role but leaves the network configuration for comcam-archiver.cp.lsst.org in place until that host can be physically removed.


glennmorris (approved):

badenerb (approved):
Lgtm

PR #1752, opened 2025-05-10
(Puppetfile) Update to official version of foreman-proxy

badenerb:


jhoblitt (commented):
I'm a bit nervous about this as I believe this won't work with EL7... I realize I keep saying we don't support EL7 anymore...

badenerb:
> I'm a bit nervous about this as I believe this won't work with EL7... I realize I keep saying we don't support EL7 anymore...

Good call, I will keep it on the branch then!

badenerb:
Keeping this one on our branch as we are not sure if it is EL& Supported

PR #1753, opened 2025-05-12
(hierdata/common) enable udev log debug

dtapiacl:
Addressing Michaels request from ticket: https://rubinobs.atlassian.net/browse/IT-6133

### DIFF
--- a/hieradata/role/fiberspec.yaml
+++ b/hieradata/role/fiberspec.yaml
@@ -5,3 +5,6 @@ classes:
   - "profile::core::fiberspec"
   - "profile::core::nfsclient"
   - "profile::ts::dco"
+
+systemd::udev_log: 'debug'
+systemd::udev_reload: true
--- a/site/profile/manifests/core/fiberspec.pp
+++ b/site/profile/manifests/core/fiberspec.pp
@@ -4,7 +4,7 @@
 class profile::core::fiberspec {
   systemd::udev::rule { 'fiberspec.rules':
     rules => [
-      "SUBSYSTEM==\"usb\", ATTR{idVendor}==\"1992\", ATTR{idProduct}==\"0667\", ACTION==\"add\", GROUP=\"saluser\", MODE=\"0664\"",
+      "SUBSYSTEM==\"usb\", ATTR{idVendor}==\"1992\", ATTR{idProduct}==\"0667\", ACTION==\"add\", GROUP=\"saluser\", MODE=\"0664\", SYMLINK+=\"fiberspec\"",
     ],
   }
 }
--- a/spec/support/spec/fiberspec.rb
+++ b/spec/support/spec/fiberspec.rb
@@ -4,7 +4,7 @@ shared_examples 'fiberspec' do
   it do
     is_expected.to contain_systemd__udev__rule('fiberspec.rules').with(
       rules: [
-        'SUBSYSTEM=="usb", ATTR{idVendor}=="1992", ATTR{idProduct}=="0667", ACTION=="add", GROUP="saluser", MODE="0664"',
+        'SUBSYSTEM=="usb", ATTR{idVendor}=="1992", ATTR{idProduct}=="0667", ACTION=="add", GROUP="saluser", MODE="0664", SYMLINK+="fiberspec"',
       ]
     )
   end


site/profile/manifests/core/fiberspec.pp
```diff
@@ -1,10 +1,23 @@
...
+#   Manages Avantes Fiberspectrometer USB devices and udev configuration.
 #
 class profile::core::fiberspec {
   systemd::udev::rule { 'fiberspec.rules':
     rules => [
       "SUBSYSTEM==\"usb\", ATTR{idVendor}==\"1992\", ATTR{idProduct}==\"0667\", ACTION==\"add\", GROUP=\"saluser\", MODE=\"0664\"",
     ],
   }
+
+  file_line { 'udev_log_debug':
```
jhoblitt: Could we use `systemd::udev_log` here?

https://github.com/voxpupuli/puppet-systemd/blob/master/manifests/init.pp#L153-L154

jhoblitt (changes requested):

hieradata/role/fiberspec.yaml
```diff
@@ -4,4 +4,9 @@ classes:
   - "profile::core::docker"
   - "profile::core::fiberspec"
   - "profile::core::nfsclient"
+  - "profile::core::systemd"
```
jhoblitt: This is already included by `profile::core::common`.

hieradata/role/fiberspec.yaml
```diff
@@ -4,4 +4,9 @@ classes:
...
   - "profile::ts::dco"
+
+systemd::manage_udevd: true
```
jhoblitt: This is already set in `hieradata/common.yaml`.

hieradata/role/fiberspec.yaml
```diff
@@ -4,4 +4,9 @@ classes:
...
+systemd::udev_log: 'debug'
+systemd::udev_reload: true
```
jhoblitt, line 10: Should we set this in the `common` layer? I think it is probably the behavior we want for all nodes.

hieradata/role/fiberspec.yaml
```diff
@@ -4,4 +4,9 @@ classes:
   - "profile::core::docker"
   - "profile::core::fiberspec"
   - "profile::core::nfsclient"
+  - "profile::core::systemd"
   - "profile::ts::dco"
+
+systemd::manage_udevd: true
+systemd::udev_log: 'debug'
```
jhoblitt, line 9: Does this need to be merged into the production config?  I'm guessing that having the host on a branch for a few hours or days would be sufficient.

jhoblitt (changes requested):

hieradata/role/fiberspec.yaml
dtapiacl: I think for now its a good idea.

hieradata/role/fiberspec.yaml
dtapiacl: This was a request from Michael to debug issues on his nodes particularly  udev rule logs related messages. I would appreciate having this merged now, and I can negotiate with him to have it removed once he is satisfied with his tests, so I don't have the hosts sitting on a branch at the moment.

hieradata/role/fiberspec.yaml
dtapiacl: Moved it up on to common layer.

hieradata/common.yaml
```diff
@@ -513,6 +513,8 @@ ccs_hcu::pkgurl: "%{lookup('profile::ccs::common::pkgurl')}"
 ccs_monit::pkgurl: "%{lookup('profile::ccs::common::pkgurl')}"
 
 systemd::manage_udevd: true
+systemd::udev_log: 'debug'
```
jhoblitt: We definitely don't want to enable debug on all hosts.

jhoblitt (changes requested):

hieradata/common.yaml
dtapiacl: Copy i will remove it from common and just move it to a role level for the hosts he is just working with which are the ones using the fiberspec role. Does this sound reasonable? thanks.

dtapiacl:
I added a symlink to link the device to a stable name for the udev rule as its all related together.

jhoblitt (commented):
The PR title suggests this PR is only enabling udev debug but that's no longer the case.

After adding the device symlink, does debug still need to be enabled on the host?

PR #1754, opened 2025-05-12
(common) bump docker version to 25.0.3

dtapiacl:
Bump docker version along with related packages. 

Im testing this on test01.ls.lsst.org with ess role applied.

### DIFF
--- a/site/profile/data/common.yaml
+++ b/site/profile/data/common.yaml
@@ -128,7 +128,7 @@ profile::core::yum::lsst_ts_private::repos:
     gpgcheck: false
     target: "/etc/yum.repos.d/lsst-ts-private.repo"
 
-profile::core::docker::version: "24.0.9"
+profile::core::docker::version: "25.0.3"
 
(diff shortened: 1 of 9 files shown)


jhoblitt (changes requested):
The tests on almalinux 8 are failing because only the package names for el9 have been updated.

jhoblitt (changes requested):
The commit title `(major/9.yaml)` isn't helpful as it is ambiguous as to which hiera layer is being modified and the commit is modifying multiple layers. 
image

My recommended fix is to squash everything into a single commit.

jhoblitt (changes requested):
image

Please state in the commit title was the version is being changed too.  Once that's done, this PR is good to go.

jhoblitt (changes requested):
Please add the docker version to the changeset title then this PR is ready to merge: 
image

jhoblitt (approved):
LGTM. Note that merging this will cause docker containers to restart, so it should wait until a maint. window.

PR #1764, opened 2025-05-15
(node/comcam-archiver.cp) remove reamaining conf

dtapiacl:
Clean up any remaining configuration from comcam-archiver.cp.lsst.org


jhoblitt (approved):
Thanks for cleaning up.

PR #1765, opened 2025-05-19
(rke2) bump dev rke2 k8s 1.31.8~rke2r1

dtapiacl:
June 2025 OS Updates.


hieradata/site/dev/role/rke2agent.yaml
```diff
@@ -0,0 +1,3 @@
+---
+rke2::release_series: "1.31"
```
cbarria: this is set in the common and since there is no mayor version jump, this can be safely removed.
https://github.com/lsst-it/lsst-control/blob/ba7eef4416384e176d56273a38483a9988d171a9/hieradata/common.yaml#L582

hieradata/site/dev/role/rke2server.yaml
```diff
@@ -0,0 +1,3 @@
+---
+rke2::release_series: "1.31"
```
cbarria: same here 👍🏻

cbarria (changes requested):

hieradata/site/dev/role/rke2server.yaml
dtapiacl: Thank you for the observation makes sense to me.

cbarria (approved):
Looks Great 👍🏻

PR #1771, opened 2025-05-26
(site/dev) add Kueyen Fluent Bit forward target in Puppet config

gseriche:
- Added new  omfwd action to rsyslog::config::actions

- Target: rsyslog.fluent.kueyen.dev.lsst.org:5140 over TCP with ossl

- Matches existing configuration style for ayekan and ruka targets

- Ensures logs are forwarded to Fluent Bit in the Kueyen environment


badenerb (approved):
LGTM

PR #1772, opened 2025-05-27
(role/hexrot) bump hexrot package versions

dtapiacl:
https://rubinobs.atlassian.net/browse/IT-6192


badenerb (approved):
Small bumps lgtm

PR #1780, opened 2025-06-02
(node/hvac01.cp) remove el7 host from repo

dtapiacl:
Entries in IPAM and IPA are also cleaned up.


cbarria (approved):
Thanks for cleaning up 🤝

PR #1781, opened 2025-06-02
(node/m1m3-dev.cp) remove el7 host from repo

dtapiacl:
There is already an EL9 instance of this VM running in production.


cbarria (approved):
Thanks for cleaning! 🤝

PR #1786, opened 2025-06-05
(site/dev) Add Kueyen Fluent Bit forward target in Puppet config

gseriche:
- Added new  omfwd action to rsyslog::config::actions
- Matches existing configuration style for ayekan and ruka targets
- Ensures logs are forwarded to Fluent Bit in the Kueyen environment


cbarria (approved):
Looks Great 👍🏻

PR #1787, opened 2025-06-06
(cluster/rancher) migrate rancher.ls RKE2

cbarria:
Migrate rancher.ls to RKE2 and clean up old config for RKE1.

_when `cp` gets migrated, we can move this config to a higher hierarchy role as its the same for each site._


dtapiacl (approved):
looks great.

PR #1791, opened 2025-06-10
(role/dm-bastion) remove role from repository

dtapiacl:
Clean up role that was tied to now retired host: dm-bastion01.cp.lsst.org.


badenerb (approved):
LGTM

PR #1793, opened 2025-06-10
(role/hexrot) Update hexrot role with bump

badenerb:


dtapiacl (approved):
Looks great. These are the versions requested correct ?

PR #1818, opened 2025-07-08
(node/tel-hw1.tu) update nfs mount point

dtapiacl:
Per Michaels request on ticket: https://rubinobs.atlassian.net/browse/IT-6357


gseriche (approved):
lgtm

PR #1819, opened 2025-07-09
(Puppetfile) Bump puppet archive to include 8.x.x

badenerb:


jhoblitt (changes requested):
This appears to be blocked by `lsst/ccs_hcu`.

PR #1825, opened 2025-07-15
(nodes/observer01) remove powertop from observer workstation

badenerb:


jhoblitt (commented):
I'm OK with this change but what problem does it resolve?

badenerb:
> I'm OK with this change but what problem does it resolve?

There are two things this fixes:

1. Mouse input delay. In order to save power, PowerTop disables a lot of USB power including mouse, which causes the user to have to shake the mouse awake every two seconds the mouse isn't being used. 
2. Monitor falling asleep. The observers often have the machine on and watching dashboards while doing work on their laptop, so they don't want the monitor to fall asleep from a lack of input- proper training about locking the screen when away from keyboard will need to happen.

I have done this with a few of the CCS Desktop machines already, these machines just weren't ready yet.

cbarria (approved):
a bit more info in the description would have avoid the question, but changes look great 👍🏻

jhoblitt (approved):
Thank you for the explanation.

PR #1830, opened 2025-07-18
(observer01.cp.lsst.org) Update network interface name

badenerb:
I had to reinstall Linux on a machine and it renamed the network interface; causing some glitches with Foreman.


dtapiacl (approved):
LGTM.

PR #1832, opened 2025-07-19
(observerstation01.cp.lsst.org) Add host config and tests

badenerb:


hieradata/node/observerstation01.cp.lsst.org.yaml
```diff
@@ -0,0 +1,17 @@
+---
+powertop::ensure: "absent"
+
+nm::connections:
```
jhoblitt, line 4: It isn't defaulting to dhcp on all physical interfaces?

jhoblitt (commented):
Is an nm configuration needed at all?

badenerb:
> Is an nm configuration needed at all?


Yes, there is a bug with this PC and other machines with the same hardware (I have no idea why) that if I don't do a nm connection on it, it will be sent to a boot loop after Puppet runs.

dtapiacl (approved):
Make sure to add these nodes on IPAM.

PR #1833, opened 2025-07-22
(node/auxtel-fp01.ls) mv /ccs-data -> /data

jhoblitt:
The mount point and NFS export path have both changed.


badenerb (approved):
Lgtm, if CI Passes

PR #1842, opened 2025-07-24
(Puppetfile) Bump CCS HCU to 3.8.1

badenerb:


jhoblitt (commented):
Would we really want switch to a -rc0 release? If not, why was review requested on a draft PR?

PR #1854, opened 2025-08-11
(role/allsky-cam) update allskycam role

dtapiacl:
-  Dimm mount no longer needed.
-  awscli2 package included it was manually installed before.


gseriche (approved):
lgtm

PR #1864, opened 2025-08-11
(observerstation02.cp.lsst.org) Add host config and tests

badenerb:


dtapiacl (approved):
LGTM. I would have, however, preferred observer-station02.cp.lsst.org as the hostname but that's personal preference.

PR #1869, opened 2025-08-12
(Puppetfile) Switch Foreman Proxy to official version

badenerb:
The commented PRs that @jhoblitt opened have been merged into the official Foreman Proxy module, for that reason, I would like to get us back on the official module.


jhoblitt:
I don't think the latest major version is compatible with our current production version: https://github.com/theforeman/puppet-foreman_proxy?tab=readme-ov-file#compatibility or operating system: https://github.com/theforeman/puppet-foreman_proxy/blob/master/metadata.json#L60-L71

badenerb:
> I don't think the latest major version is compatible with our current production version: https://github.com/theforeman/puppet-foreman_proxy?tab=readme-ov-file#compatibility or operating system: https://github.com/theforeman/puppet-foreman_proxy/blob/master/metadata.json#L60-L71

Per Josh's comment, I will close this PR

PR #1890, opened 2025-09-12
(node/lsstcam-{pcs,vw}02.cp) fix nm::connections

glennmorris:


badenerb (approved):
LGTM, thanks for catching those!

jhoblitt (approved):

PR #1895, opened 2025-09-16
(Puppetfile) Switch simp-mate to local fork

badenerb:
The official Puppet Module for simp-mate has not had a release since 2023, so in that case I would like to move to our own fork, that is up to date with the module.


jhoblitt (changes requested):
The PR title should state what is being switched to a local fork.

jhoblitt (approved):

PR #1906, opened 2025-09-30
(cluster/lukay) convert rke1 -> rke2

cbarria:
- Migrate to RKE2

### DIFF
--- a/hieradata/cluster/lukay.yaml
+++ b/hieradata/cluster/lukay.yaml
@@ -2,3 +2,7 @@
 clustershell::groupmembers:
   lukay: {group: "lukay", member: "lukay[01-04]"}
 tuned::active_profile: "latency-performance"
+nm::conf:
+  device:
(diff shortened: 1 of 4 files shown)


dtapiacl (approved):
LGTM.

dtapiacl (approved):

gseriche (approved):
lgtm

dtapiacl (commented):
1. Question: Where is the network configuration for the cluster ?

jhoblitt (changes requested):
For a 4 (or 5) node cluster, I think all nodes should be control plane. Now also seems like a good time to migrate nfs2.cp back to a lukay node.

As part of any baremetal k8s cluster rebuild, in the interested of reliability and consistency, the networking should be converted to use a bonded trunk with the fqdn IP on bridge interface -- even if there is only a single vlan on the trunk.

PR #1913, opened 2025-10-07
(node/perfsonar1.ls) upgrade base perfsonar to alma9

julioconstanzo:
Upgrade base perfsonar to alma9 and add routes for the perfsonar lhn environment.


dtapiacl (commented):
There is an error on the network configuration of perfsonar1.ls.lsst.org.yaml, there is a mismatch on the name of the interfaces.

dtapiacl (approved):
LGTM.

PR #1915, opened 2025-10-07
(node/lsstcam-db01.cp) install ccs_database cronjob

glennmorris:


tony-johnson (approved):
Looks good to me. Am I right in assuming that this does not define the cron job, but expects it to exist in /opt/lsst/ccs/prod/bin

jhoblitt (approved):

PR #1918, opened 2025-10-08
(Puppetfile) Switch to local fork for puppet-tftp

badenerb:
Switching to a local fork because the official version dropped EL8


gseriche (approved):
lgtm

PR #1937, opened 2025-10-21
(role/hexrot) Update hexrot role with bump to Python v3.13 and modules

badenerb:
Updating the modules in the files as requested from Te-Wei.


jhoblitt (commented):
The PR title is kind could be more helpful and state what is being bumped.

jhoblitt (changes requested):
The GHA is complaining about the length of the PR title: 
image

jhoblitt (approved):
LGTM

PR #1948, opened 2025-10-23
(common) add 139.229.161.96/28 vlan 1131

igonzalezcl:
Allow network 139.229.161.96/28 to contact tang.


jhoblitt (changes requested):
Creating a new ipset on its own isn't going to allow access to tang as the existing nftables rules are referencing the `cpfde` ipset.

dtapiacl (approved):
LGTM.

jhoblitt (changes requested):
The `merken` ipset being introduced in this PR isn't used and should be removed.

csilva-cl:
> The `merken` ipset being introduced in this PR isn't used and should be removed.

merken will be re-deployed with enc, is that what you meant ?

jhoblitt (approved):
LGTM.

PR #1958, opened 2025-11-03
(common.yaml) clean up cbarria_b backdoor access

dtapiacl:
Cleans up backdoor access to Rubin Infrastructure.


jhoblitt (commented):
I think we should probably add the `cbarria` account to the purge list as well.

jhoblitt (approved):
LGTM.

PR #1969, opened 2025-11-05
(role/hexrot) Update hexrot roles various packages

badenerb:
Also included updates from this ticket: IT-6562


igonzalezcl (approved):
LGTM

PR #1976, opened 2025-11-10
(node/auxtel-dc1.cp.lsst.org) add vm host

dtapiacl:
Puppetizes auxtel-dc1.cp.lsst.org VM network configuration. This VM will be tested by Tony. If it behaves well, it will replace the auxtel-dc01.cp.lsst.org Dell server. 

Device was added on IPAM.


gseriche (approved):
lgtm

PR #1979, opened 2025-11-11
(role/moss) add new role and tests

dtapiacl:
This takes care of ticket: https://rubinobs.atlassian.net/browse/IT-6566


gseriche (approved):
lgtm

PR #1980, opened 2025-11-11
(node/tel-hw1.cp.lsst.org) add azar2.cp replacement node

dtapiacl:
Create a VM to replace azar2.cp.lsst.org at the summit.


gseriche (approved):
lgtm

PR #1987, opened 2025-11-13
(node/ipa1.cp.lsst.org) enable restic backups of ipa

jhoblitt:
As this is an offline backup, which takes down freeipa, we only want to enable it on a single node at a time.


gseriche (approved):
lgtm

PR #1988, opened 2025-11-17
(rke2) bump CP rke2 k8s 1.32.9~rke2r1

badenerb:
Lukay is omitted because it is currently on a branch. I will make the updates to the branch it is on when CP updates begin.


dtapiacl (approved):
LGTM.

PR #1989, opened 2025-11-17
(Puppetfile) Update Foreman Ref to most recent

badenerb:


jhoblitt (changes requested):
The foreman module is updated in https://github.com/lsst-it/lsst-control/pull/1928, which is live on dev, and ls, and will hopefully complete rollout to all sites in the near future. I'd like to close this PR in favor of that one.

PR #1991, opened 2025-11-19
(node/lsstcam-daq-mgt.cp) decomission lsstcam-daq-mgt.cp

dtapiacl:
This server will no longer be of use as it was replaced by lsstcam-daq.cp.lsst.org.


badenerb (approved):
LGTM!!

PR #2015, opened 2025-12-12
(*/rke) rm rke files

badenerb:
Our environment no longer has any rke k8 clusters, so I am removing these files to cleanup.


badenerb:
Blocked while we wait for branches to be merged for other clusters

PR #2021, opened 2026-01-05
(hierdata/node) add static ip lsstcam-aio*.cp

dtapiacl:
Addresses ticket: https://rubinobs.atlassian.net/browse/IT-6613


gseriche (approved):
lgtm

PR #2041, opened 2026-02-03
(profile::core::podman::prune) add podman prune profile

dtapiacl:
Adds a daily cron job to prune unused Podman resources older than 14 days.

### DIFF
--- a/site/profile/manifests/core/podman/prune.pp
+++ b/site/profile/manifests/core/podman/prune.pp
@@ -1,0 +1,16 @@
+# @summary
+#   Enable periodic cleanup of unused podman resources
+#
+class profile::core::podman::prune {
+  cron::job { 'podman_prune':
+    minute      => '0',
+    hour        => '16',
+    date        => '*',
+    month       => '*',
+    weekday     => '*',
+    user        => 'root',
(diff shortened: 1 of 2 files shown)


site/profile/manifests/core/podman/prune.pp
```diff
@@ -0,0 +1,16 @@
...
+#
+class profile::core::podman::prune {
+  cron::job { 'podman_prune':
+    minute      => '0',
+    hour        => '16',
+    date        => '*',
+    month       => '*',
+    weekday     => '*',
+    user        => 'root',
+    command     => 'systemd-cat -t podman-prune podman system prune -a --filter "until=$((90*24))h" --force',
```
jhoblitt: 90 days seems like a lot. That only allows ~22 branch pushes per day before the 2048 lock limit would be hit again.  I'd suggest a reducing this to a couple of weeks.

jhoblitt (changes requested):

site/profile/manifests/core/podman/prune.pp
dtapiacl: Thank you for the feedback. I will update it.

jhoblitt (approved):
LGTM.

PR #2043, opened 2026-02-03
(role/niagara) add hvac restic backup

dtapiacl:
1. Adds the ability to perform restic backups on some bare metal hosts. 
2. `profile::core:restic` was moved inside `profile::core::common` to discontinue the direct inclusion of `restic` on other roles.

### DIFF
--- a/spec/hosts/roles/niagara_spec.rb
+++ b/spec/hosts/roles/niagara_spec.rb
@@ -22,6 +22,20 @@ describe "#{role} role" do
           it { is_expected.to compile.with_all_deps }
 
           include_examples('common', os_facts:, site:)
+
+          it { is_expected.to contain_sudo__conf('niagara_sudoers').with_content('%niagara ALL=(ALL) NOPASSWD: /usr/bin/niagaradctl') }
+
+          it { is_expected.to contain_class('restic') }
+
+          it {
+            is_expected.to contain_restic__repository('awsrepo').with(
+              backup_path: ['/etc/mosquitto', '/opt/Niagara'],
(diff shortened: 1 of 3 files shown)


csilva-cl (approved):
looks good

csilva-cl:
I would remove the "test" from the PR

jhoblitt (changes requested):
This looks reasonable but needs unit tests.

spec/hosts/roles/niagara_spec.rb
```diff
@@ -22,6 +22,20 @@
           it { is_expected.to compile.with_all_deps }
 
           include_examples('common', os_facts:, site:)
+
+          it { is_expected.to contain_sudo__conf('niagara_sudoers').with_content('%niagara ALL=(ALL) NOPASSWD: /usr/bin/niagaradctl') }
+
+          it { is_expected.to contain_class('profile::core::restic') }
```
jhoblitt: It would be better to check for the inclusion of the `restic` class. This may seem a bit confusing but the goal of the unit tests is to inspect resources that are created in the catalog and not to mirror the structure of the roles/profile pattern. So while there is a class container named `profile::core::restic` what we actually care about in this case is that the `restic` class is included.  Checking only for the profile doesn't tell us if the `restic` class is included.  Generally, the unit tests should probe directly for resources and not classes but in the case of external modules, checking for the class is OK.

jhoblitt (changes requested):

jhoblitt (approved):
LGTM.

PR #2044, opened 2026-02-03
(role/foreman) add podman prune profile

dtapiacl:


hieradata/role/foreman.yaml
```diff
@@ -7,6 +7,7 @@ classes:
   - "profile::core::docker::prune"
   - "profile::core::foreman"
   - "profile::core::foreman::fog_hack"
+  - "profile::core::podman::prune"
```
jhoblitt, line 10: The cronjob will fail on foreman.tu and foreman.cp as they haven't been updated to el9/foreman 3.16 yet. However, it shouldn't cause an error to be reported from a puppet agent run, so I'm OK with it.

jhoblitt (approved):

PR #2045, opened 2026-02-04
(Puppetfile) bump ccs_monit to 3.0.0

glennmorris:
This should suppress some bogus warnings.


tony-johnson (approved):
Looks good to me, and would be good to suppress the bogus messages.

PR #2048, opened 2026-02-05
(observer0*) rm old observers configs and tests

badenerb:
These two machines are being used with a better FQDN and a new location, so I am cleaning these up. Foreman, IPA, and IPAM entries have all been deleted.


dtapiacl (approved):
LGTM. Thanks for cleaning up.

PR #2052, opened 2026-02-10
(role/openvpnas) add vpn-electric group

badenerb:


dtapiacl (approved):
Looks great, thank you for testing it on DEV first.

PR #2057, opened 2026-02-13
(role/ipareplica) change ipa restic backups to role level

badenerb:


spec/hosts/roles/ipareplica_spec.rb
```diff
@@ -101,6 +101,20 @@
             end
           end
         end # host
+
+        it do
```
jhoblitt: I suspect this block needs to be moved inside of the loop over hosts.

jhoblitt (approved):
LGTM. Thanks for fixing the tests.

PR #2101, opened 2026-02-26
(role/dnscache) add dnssec exception for lsst.local

badenerb:
This fixes the problem of not being able to reach lsst.local on our networks. This adds lsst.local as an exception to our DNSSec policy of `auto` (turned on), by adding a DNS class and a test for it.

### DIFF
--- a/hieradata/role/dnscache.yaml
+++ b/hieradata/role/dnscache.yaml
@@ -5,6 +5,9 @@ classes:
 dns::forward: "only"
 dns::listen_on_v6: false
 dns::sysconfig_startup_options: "-4"
+dns::dnssec_validation: "auto"
+dns::additional_options:
+  "validate-except": '{ "lsst.local"; }'
 dns::additional_directives:
   - "logging {"
   - "    channel queries_syslog {"
--- a/spec/hosts/roles/dnscache_spec.rb
+++ b/spec/hosts/roles/dnscache_spec.rb
@@ -29,7 +29,11 @@ describe "#{role} role" do
                 '1.0.0.1',
                 '1.1.1.1',
                 '8.8.8.8',
-              ]
+              ],
+              dnssec_validation: 'auto',
+              additional_options: {
+                'validate-except' => '{ "lsst.local"; }',
+              }
             )
           end
 


site/profile/manifests/core/dns.pp
```diff
@@ -0,0 +1,27 @@
...
+
+  exec { 'add_lsst_local_nta':
+    command     => '/usr/sbin/rndc nta -lifetime 604800 lsst.local',
+    path        => ['/usr/sbin', '/usr/bin', '/bin'],
+    require     => [Service['named'], Exec['remove_lsst_local_nta']],
+    refreshonly => true,
+  }
+
+  # Force the add exec to run every puppet run by refreshing it.
+  notify { 'renew_lsst_local_nta':
```
jhoblitt: Per the slack discussion, using the puppet agent as a cron daemon isn't a good design pattern.  However, lets say we did want two `exec`s to run on every agent run with ordering.  The `notify` resource wouldn't be necessary as the `exec` will, by default, trigger.  The `exec`s could have explicit resource ordering or one `exec` could explicitly trigger a refresh only `exec`. E.g.:

```puppet
  exec { 'remove_lsst_local_nta':
    command => '/usr/sbin/rndc nta -remove lsst.local',
    returns => [0, 1],
    require => Service['named'],
    notify  => Exec['add_lsst_local_nta'],
  }

  exec { 'add_lsst_local_nta':
    command     => '/usr/sbin/rndc nta -lifetime 604800 lsst.local',
    require     => Service['named'],
    refreshonly => true,
  }
```

site/profile/manifests/core/dns.pp
```diff
@@ -0,0 +1,27 @@
+# @summary
+#   Disable DNSSec for the lsst.local zone
+#
+class profile::core::dns {
+  include dns
+
+  exec { 'remove_lsst_local_nta':
+    command => '/usr/sbin/rndc nta -remove lsst.local',
```
jhoblitt: It looks like there is an `additional_options` param that could, in theory, create:

```
options {
    dnssec-validation auto;

    validate-except {
        "lsst.local";
    };
}; 
```

but I'm unsure if it would format the block correctly.

https://github.com/theforeman/puppet-dns/blob/master/manifests/init.pp#L194
https://github.com/theforeman/puppet-dns/blob/d77a34c661764091d91e2e1e64ac686a62a63c35/templates/options.conf.epp#L50

It might work if the "value" passed was something like `{ "lsst.local"; }`.

site/profile/manifests/core/dns.pp
badenerb: I resolved this by changing how I approached the issue, I used @jhoblitt 's solution with the `additional_options` instead of scary `exec`

site/profile/manifests/core/dns.pp
badenerb: I ended up using this exact approach slightly modified

site/profile/manifests/core/dns.pp
```diff
@@ -0,0 +1,11 @@
+# @summary
+#   Disable DNSSEC validation for the lsst.local zone
+#
+class profile::core::dns {
+  class { 'dns':
+    dnssec_validation  => 'auto',
+    additional_options => {
```
jhoblitt: Did this not work via hiera?

badenerb: Yes it does, modified my code accordingly.

dtapiacl (approved):
LGTM.

jhoblitt (approved):
Looks great now. Thanks for sticking with this.

PR #2108, opened 2026-03-11
(node/comcam-hcu03.cp) add missing packages

badenerb:


jhoblitt (approved):
This LGTM to me but it would be good if @glennmorris could take a look.

glennmorris (approved):
Thanks; LGTM.

PR #2122, opened 2026-03-12
(node/foreman.ls) add MACHI|TOKI|LONKO subnets

jhoblitt:


dtapiacl (approved):
Subnetting looks good to me

PR #2124, opened 2026-03-14
(cp/role/ccs-mcm) migrate ccs sudo setting here from host-specific

glennmorris:
So that it also applies to lsstcam-mcm1.cp.


tony-johnson (approved):
Looks good to me

PR #2134, opened 2026-03-26
(toqui) replace name from toki to toqui

dtapiacl:
- Updates the cluster name to its correct format.
- IPAM was updated to reflect this along with foreman LS settings.


badenerb (changes requested):
Please update IPAM

csilva-cl (approved):
looks reasonable.

PR #2136, opened 2026-03-27
(role/puppetdb) bump java args resources

dtapiacl:
Bumps JVM heap limit from 1GB to 3GB as it's too small for the workload it handles.

This was causing OOM and puppet db to crash thus generating big amounts of logs.


csilva-cl (approved):
looks good

PR #2151, opened 2026-04-21
(hierdata/cluster) update cluster shell group

dtapiacl:
This PR updates the cluster shell group members from 10 to 11 which is the amount of Manke nodes that we have.


badenerb (approved):
LGTM if CI Passes

PR #2152, opened 2026-04-21
(node/auxtel-tcbp01.cp.lsst.org) provision host and add test

badenerb:


dtapiacl (approved):
LGTM. But please check connectivity in that private subnet before merging.

PR #2154, opened 2026-04-22
(hieradata/foreman.cp) remove pdu from summit foreman

julioconstanzo:
remove pdu from summit foreman


dtapiacl (approved):
LGTM. Merge when CI passes.

PR #2157, opened 2026-04-27
(cluster/machi) add toqui agent nodes network

dtapiacl:
- Removes br2141 from cluster/machi.yaml so it can be managed per-node. 
  - Regular machi nodes (variant=2015s) are unaffected. br2141 DHCP config is already defined in cluster/machi/variant/2015s.yaml                                                                        
  - Adds static IP assignment on br2141 for toqui nodes (variant=1115hs) without creating per-node
  files.


csilva-cl (approved):
looks reasonable

PR #2158, opened 2026-04-27
(role/daq-mgt) add rce SSH config for DAQ network

jgthayer:
As described in https://rubinobs.atlassian.net/browse/IT-6906, this adds the RCE ssh key to the rce users .ssh file and sets the ssh config to use it when communicating with the RCEs on the private DAQ network.

### DIFF
--- a/hieradata/role/daq-mgt.yaml
+++ b/hieradata/role/daq-mgt.yaml
@@ -55,3 +55,28 @@ accounts::user_list:
     <<: *daq_user
     uid: 62003
     gid: 62003
+
+# SSH key and config for rce service account to connect to RCE embedded
+# systems on the DAQ network (192.168.100.0/24).
+# The .ssh directory is created by the accounts module (managehome: true).
+# The private key content is in lsst-puppet-hiera-private/role/daq-mgt.eyaml.
+files:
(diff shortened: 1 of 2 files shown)


glennmorris (approved):
Looks ok to me.
I recommend requesting reviews from `lsst-it/cl-puppet` in future.

jgthayer:
@glennmorris Apologies! I didn't even know about it. I guess I'm punishing you for being the only person that's ever responded to me! ;-)

jgthayer:
It looks like I'm not allowed to merge anymore, either. I'll tag them now @lsst-it/cl-puppet and maybe they'll merge for me too.

dtapiacl (commented):
Was this tested on the host and works ? @jgthayer  We can use foreman.cp to this by moving the host to its branch.

jgthayer:
No, it wasn't. Testing on the branch sounds like a prudent idea. It looks like the PR https://github.com/lsst-it/lsst-puppet-hiera-private/pull/195 was merged to production, so does that mean that change will be present? I've never been sure how these repos interact!

dtapiacl (approved):
LGTM. Thanks for testing this.

PR #2177, opened 2026-05-08
(profile::ccs::file_transfer) remove -s3nd script versions

glennmorris:
These are no longer used.


tony-johnson (commented):
As usual my language was not sufficiently precise, we still need send-s3nd (it is called by the other non -s3nd scripts)

glennmorris:
> As usual my language was not sufficiently precise, we still need send-s3nd (it is called by the other non -s3nd scripts)

Restored.

tony-johnson (approved):
Looks good to me.

dtapiacl:
Looks greats

dtapiacl (approved):

PR #2178, opened 2026-05-08
(site/cp/cluster/lsstcam-ccs) update clustershell::groupmembers

glennmorris:


tony-johnson (commented):
I don't think we should remove dc-new yet, since we may have told people to use it (and it may even be documented). Adding dc-old and leaving dc-new as a synonym for dc would be good. We also added lsstcam-ser (shutter extraction rail) (an uno) although it does not seem to be online right now.

glennmorris:
> I don't think we should remove dc-new yet, since we may have told people to use it (and it may even be documented). Adding dc-old and leaving dc-new as a synonym for dc would be good. We also added lsstcam-ser (shutter extraction rail) (an uno) although it does not seem to be online right now.

OK, I made dc-new the same as dc.
I had already added lsstcam-ser.

dtapiacl (approved):
Looks good to me.

PR #2188, opened 2026-05-20
(node/ats-*) add missing network configs and tests

badenerb:
Upon reboot the 3 hosts do not bring up their network interfaces on their own. These three network configurations should fix that bug allowing them to be connected to network on boot.


dtapiacl (approved):
Looks reasonable.

PR #2197, opened 2026-06-15
(profile::pi::gphoto2) add gphoto2 RPM

mareuter:


dtapiacl (commented):
LGTM. @mareuter Did you test this on the host as well ?. You can do this by moving the host on to the branch on Foreman and running puppet.

mareuter:
Yes, I had already switched the machine foreman entry over to the branch environment. Baden helped me check the catalog on the machine to see that it picked up the change.

dtapiacl (approved):
LGTM.

PR #2201, opened 2026-06-19
(common) add ccs_sal::rpms, to work around an error

glennmorris:
Without this we get the following error, no idea why:
  Error while evaluating a Method call, Class[Ccs_sal]: expects a value
  for parameter 'rpms'


badenerb (approved):
lgtm

PR #2203, opened 2026-06-23
(common) remove ccs_sal::rpms, no longer used

glennmorris:
This was only added as a workaround for an unexplained puppet error. It now seems that the error was due to foreman serving a stale version of cp_production due to an internal problem.


dtapiacl (approved):
LGTM.

PR #2204, opened 2026-06-26
(profile::core::fiberspec) change FiberSpec udev to use numeric GID

mareuter:


dtapiacl (approved):
Looks good to me. Tested and working ?

mareuter:
Yes. We applied the branch, rebooted the machine we applied it to and the correct group got assigned right away.

PR #2205, opened 2026-06-29
(node/comcam-db01.cp) do not manage the database service state

glennmorris:
Since comcam-db02 is now the main database server, we do not want the mariadb service to be automatically started on db01.


dtapiacl (approved):
Looks good to me.

PR #2224, opened 2026-07-08
(core/rucio) pin xrootd version and depednecies

badenerb:
Regarding this ticket: IT-6984

Tested and working as expected on Rucio01.ls.lsst.org


dtapiacl (approved):
Looks great ! Thank you for fixing this and adding tests.

